PR #30, opened 2021-02-21
Fix table inconsistency

jaydhulia:


ahmedhamodi (changes requested):
Since it's an array of objects, I think object_ids makes a lot more sense. If this is an issue in our create script, can you change in there instead?

jaydhulia:
Leaving this fix for now. If the change needs to be done in the create script, then it will also need to be updated in mock data.

PR #31, opened 2021-02-21
Create base model for useful methods across all our tables.

jaydhulia:
- Create a base model table, so that we can have common methods across all our models in here (such as converting to dict etc.)


ahmedhamodi (approved):
lgtm! thanks for doing this

PR #33, opened 2021-02-23
Integration of Tornado <> DB Client

jaydhulia:
- Added GET and CREATE for the tables

### DIFF
--- a/.run/dev-viv.run.xml
+++ b/.run/dev-viv.run.xml
@@ -1,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="dev-viv" type="PythonConfigurationType" factoryName="Python">
+    <module name="cdc-backend" />
+    <option name="INTERPRETER_OPTIONS" value="" />
+    <option name="PARENT_ENVS" value="true" />
+    <envs>
+      <env name="PYTHONUNBUFFERED" value="1" />
+      <env name="CONFIG_PATH" value="../configs/dev-config.yaml" />
+    </envs>
+    <option name="SDK_HOME" value="C:\Users\vivia\Documents\Blueprint\cdc-backend\env\Scripts\python.exe" />
+    <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/app" />
+    <option name="IS_MODULE_SDK" value="false" />
+    <option name="ADD_CONTENT_ROOTS" value="true" />
+    <option name="ADD_SOURCE_ROOTS" value="true" />
+    <EXTENSION ID="PythonCoverageRunConfigurationExtension" runner="coverage.py" />
+    <option name="SCRIPT_NAME" value="$PROJECT_DIR$/app/__main__.py" />
+    <option name="PARAMETERS" value="" />
+    <option name="SHOW_COMMAND_LINE" value="false" />
+    <option name="EMULATE_TERMINAL" value="false" />
+    <option name="MODULE_MODE" value="false" />
+    <option name="REDIRECT_INPUT" value="false" />
+    <option name="INPUT_FILE" value="" />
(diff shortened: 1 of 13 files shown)


app/models/db_client.py
```diff
@@ -27,7 +27,7 @@ def create_entity(data):
     session.close()
 
     # return the id to the caller
-    return data.id
+    return data
```
ahmedhamodi, line 30: I'd prefer to return ID here. What's the use case for returning the whole payload?

jaydhulia: Couple of reasons for that:
1. The frontend will need the entire payload to render the page; so if the frontend creates a scenario and just gets the ID, the frontend will need to make 2 requests to render the page, first for the POST, and then a GET for the scenario
2. I don't see a reason to NOT return the newly created object. There are certain things that will only be populated once the object is created (`id` is just one example, but other things like `scene_ids`, `is_published` etc. are not part of the initial payload, but are part of the newly created object).

Is there an explicit reason for only wanting the id to be returned?

ahmedhamodi: I don't think #1 is always true. If I am creating a new asset, I upload an image and give a name/description and then ship it off - I don't want the whole payload back. I will, at a different time, request the asset.

In fact, isn't the flow I described above the way it's going to work when creating almost everything? You won't immediately use it, we will have a different flow for creating stuff.

The main reason to not give back the whole payload is that it can increase the payload size and slow down the endpoints. In fact, we only want to return a status code on a POST request imo.

jaydhulia: You pointed out the only instance when 1 isn't true. In other cases, it will be true. Create a new scenario? -> GET scenario to render it. Create a new scene? -> GET scene to render it. Create a new object? -> GET object to render it on the scene. Add text to a scene? GET text to render it and so on...

It will be significantly slower to do 2 requests than returning the entire payload. The entire payload is just a JSON obj with less than 20 keys.

ahmedhamodi: I still don't think that's true. It's hard to communicate this when we haven't finalized admin design but my idea of the "scenario/scene" editors is that you just have a dropdown of objects, text, and assets to use. Even with making scenes and scenarios, you should be able to create one independent of immediately loading it into the scenario.

But fair point, the payload isn't massive so perhaps it's not a big deal.

.run/dev-viv.run.xml
```diff
@@ -0,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
```
ahmedhamodi, line 1: Should we be committing this file?

ahmedhamodi (approved):
LGTM except the question about `dev-viv.run.xml`. Doesn't feel like it should be included

.run/dev-viv.run.xml
jaydhulia: Yeah it is okay to include; it is just the Pycharm run configuration so it will load the Pycharm configurations properly if someone opens the project in Pycharm. It will set everything properly, and just need to update the python path to get it to run.
The `dev.run.xml` file works for Mac/Linux, this one is for Windows (Vivian specifically).

ahmedhamodi: Let's pick a better name in the future since both Dhruvin and Amolik will be using Windows too!

jaydhulia: While the Max/Linux files are the same for everyone, the Windows file seem to be different per person. Hence the name :)

jaydhulia: I will investigate if I can make a generic windows one though

ahmedhamodi: Ew windows why you like this. But okay, if that is possible that would be amazing!

PR #41, opened 2021-02-26
Supporting delete in db client

ahmedhamodi:
If unable to delete, error message propagated:
image

If successful (this is just hardcoded in tornado client side, can change):
image

If deleting a non-existent id (example, asset id 5 after it was already deleted):
image

### DIFF
--- a/app/library/postgres.py
+++ b/app/library/postgres.py
@@ -76,12 +76,12 @@ async def get_asset_from_postgres(asset_id: str):
 
 async def delete_asset_from_postgres(asset_id: str):
     # TODO: delete from postgres
-    is_deleted = delete_asset(asset_id)
+    try:
+        delete_asset(asset_id)
+    except Exception as e:
+        raise e
 
-    if is_deleted:
-        response = {"message": "deleted successfully"}
-    else:
-        response = {"message": "error deleting asset with id " + asset_id}
+    response = {"message": "deleted successfully"}
     return response
 
 
--- a/app/library/postgres.py
+++ b/app/library/postgres.py
@@ -1,6 +1,7 @@
 from models.asset import Asset
 from models.db_client import (
     create_entity,
+    delete_asset,
     get_asset,
     get_object,
     get_scenario,
@@ -75,9 +76,13 @@ async def get_asset_from_postgres(asset_id: str):
 
 async def delete_asset_from_postgres(asset_id: str):
     # TODO: delete from postgres
+    is_deleted = delete_asset(asset_id)
 
-    sample_response = {"message": "deleted successfully"}
-    return sample_response
+    if is_deleted:
+        response = {"message": "deleted successfully"}
+    else:
+        response = {"message": "error deleting asset with id " + asset_id}
+    return response
 
 
 async def update_asset_from_postgres(asset_id: str, data: dict):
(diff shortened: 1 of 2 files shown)


drbalar (approved):
LG

jaydhulia:
> Todo:
> 
> * explore exception error and possibly propagating that up instead of a boolean

Are you going to be adding the TODO to this PR and then review, or is that a TODO for a later date? @ahmedhamodi

ahmedhamodi:
I was going to do the todo in this PR probably tomorrow (today since it's after midnight).

app/library/postgres.py
```diff
@@ -72,9 +73,13 @@ async def get_asset_from_postgres(asset_id: str):
 
 async def delete_asset_from_postgres(asset_id: str):
     # TODO: delete from postgres
+    try:
+        delete_asset(asset_id)
+    except Exception as e:
+        raise e
```
jaydhulia, line 79: Let's do `delete_asset(asset_id)` just as is here, so we don't slow down the response by catching the exception and then propagating it.
Alternatively, if we want to make this more DCC friendly, we should catch it and turn it human readable:
```
try:
        delete_asset(asset_id)
    except psycopg2.errors.ForeignKeyViolation as e:
        raise AssertionError("One or more scenes are referencing this asset, please delete the scene before deleting this asset")
```

app/library/postgres.py
```diff
@@ -72,9 +73,13 @@ async def get_asset_from_postgres(asset_id: str):
...
 
-    sample_response = {"message": "deleted successfully"}
-    return sample_response
+    response = {"message": "deleted successfully"}
```
jaydhulia, line 81: If an asset doesn't exist in the table, this method still returns `{"message": "deleted successfully"}`. Should we do a check to see if the asset exists first? Or is there a better way of distinguishing between non-valid asset IDs and valid asset IDs that are deleted?

app/models/db_client.py
```diff
@@ -46,6 +46,18 @@ def get_asset(id):
     return asset
 
 
+# TODO: move session creating and closing out of these individual calls
+def delete_asset(id):
+    try:
+        session = Session()
+        session.query(Asset).filter(Asset.id == id).delete(synchronize_session="fetch")
```
jaydhulia, line 53: Out of curiousity, what does `synchronize_session="fetch"` do?

app/library/postgres.py
ahmedhamodi: We shouldn't really allow the deletion of non-valid asset IDs in the frontend UI. But either way, I agree a different error message should be returned. I'll look into that

app/models/db_client.py
ahmedhamodi: As per discussion offline:
image

Will investigate if delete needs to be synchronized or not.

app/library/postgres.py
```diff
@@ -74,10 +75,13 @@ async def get_asset_from_postgres(asset_id: str):
...
-    # TODO: delete from postgres
+    did_delete = delete_asset(asset_id)
 
-    sample_response = {"message": "deleted successfully"}
-    return sample_response
+    if did_delete:
+        response = {"message": "deleted successfully"}
+    else:
+        response = {"message": "asset id " + asset_id + " not found"}
+    return response
```
ahmedhamodi, line 84: This is just an example of how I envision the deletion being used, but this is for @jaydhulia and @vivianliu0 to implement and change as needed.

ahmedhamodi:
Note: leaving the cleaning of error message to your implementation/decision on Tornado side @jaydhulia @vivianliu0

jaydhulia (approved):

PR #49, opened 2021-03-12
Connection pooling and session concurrency

drbalar:


app/models/db_client.py
```diff
@@ -5,14 +5,13 @@
 from models.statistics import Statistics
 from models.text import Text
 
-from . import Base, Session, engine
+from . import Base, engine
 
 
 # TODO: move session creating and closing out of these individual calls
```
jaydhulia, line 11: Can we remove all the TODOs now?

app/routes/admin/asset.py
```diff
@@ -23,7 +23,7 @@ async def post(self):
             # validate body
             validate(data, schema=admin_asset_handler_body_schema)
 
-            inserted_asset = await post_asset_to_postgres(data)
+            inserted_asset = await post_asset_to_postgres(data, self.session)
```
jaydhulia, line 26: Can we name this something like `self.db_session` instead? Because we will also have user auth sessions in the future

jaydhulia (changes requested):

app/models/db_client.py
drbalar: yes lol

app/models/__init__.py
```diff
@@ -11,8 +11,16 @@
 db_url = config.get("postgres.db_url").format(
     user=user, password=password, database=database, hostname=hostname
 )
-engine = create_engine(db_url)
+engine = create_engine(db_url, pool_size=2, max_overflow=1)
```
jaydhulia, line 14: Should this be 2 and 1? I thought we were doing 25 and 5?

jaydhulia (approved):
LGTM

ahmedhamodi (approved):
LGTM! Thanks for doing this 🎉

PR #52, opened 2021-03-13
Remove text references and table

jaydhulia:
- [x] Update config
- [x] Update create table / insert data scripts
- [x] Update tornado handlers
- [x] Update DB Clients - remove text table 
- [x] Update postgres.py
- [x] Update API_Schema handlers


ahmedhamodi (commented):
Code changes look good. What level of testing did you do to make sure none of the current supported end points break?

jaydhulia:
I can't reply to your comment directly @ahmedhamodi, but basically I ran the app and went through all the saved endpoints that I had on Postman and all of them worked :) Also ran the update/insert scripts and manually checked postgres to ensure it looked good

ahmedhamodi (approved):
Great thanks for providing that context. Looks good to me!

drbalar (approved):

PR #54, opened 2021-03-26
Server side rendering - Initial PR

jaydhulia:
- [x] Serve a static folder
- [x] Error page
- [x] Include Auto-reload on changes to HTML
- [x] Handler for scenario UI
- [x] Handler for static files
- [x] Sample scene page


jaydhulia:
image
Just prints out the information for the scene and scenario for a valid scenario.

jaydhulia:
image

Invalid Scenario/Scene is a generic error for now (designers will update it :) )

app/__main__.py
```diff
@@ -51,6 +54,13 @@ def get_routes():
             AdminObjectPutHandler,
         ),
         (r"/api/admin/v1/scene/([0-9]{1,16})/duplicate", AdminSceneDuplicateHandler),
+        (
+            r"/static/(.*)",
+            UIStaticHandler,
+            dict(path=f"{os.path.dirname(__file__)}/public/static/"),
```
ahmedhamodi, line 60: Is this using our local path? When this is deployed on a server, what is this going to be?

app/routes/ui/scenario.py
```diff
@@ -0,0 +1,45 @@
...
+            scene_number_int = int(scene_number)
+            # TODO: caching will be dealt with in a later PR
+            scenario_obj: Scenario = await get_scenario_by_friendly_name_from_postgres(
+                scenario_friendly_name, self.db_session
+            )
+            if not scenario_obj.is_published:
+                raise ValueError("This Scenario is currently not accessible")
+
+            if scene_number_int >= len(scenario_obj.scene_ids) or scene_number_int < 0:
+                raise ValueError("Invalid Scene")
```
ahmedhamodi, line 25: Order of checks? Should we check invalid before is published?

Pros: if they see it's not accessible, they'll assume it exists even though it might not.

Cons: malicious users can go through and find which scene ids exist, published or not.

I'll let you decide which is best - won't block the PR for this.

ahmedhamodi (approved):
lgtm. Couple comments, I'll let you make the final decision on these.

app/routes/ui/scenario.py
jaydhulia: No, so this order is what the most secure way. If a scenario is not published, then we should leak *no* information, hence why the scene ids check is only after ensuring that the scenario is supposed to be accessible.

app/__main__.py
jaydhulia: This is the path of the file where it this file exists (__main__.py). So I am saying get the directory where __main__.py exists, and append `public/static` to it. Meaning it will always be correct, regardless of running locally or on server

PR #55, opened 2021-03-27
Jigsaw Prototype

ahmedhamodi:
Loom video showing the prototype in action:
https://www.loom.com/share/b7d24009ca044034b6c9a59337841e52

Adding Aaron as a reviewer too.


samples/images/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.png
```
jaydhulia, line 1: Let's not add these to .gitignore please; Since tornado will be used to do server-side rendering, we will have .png, .jpg, .jpeg files in here. If you need to restrict it, please restrict the .gitignore to only a single folder

ahmedhamodi: It's only in samples/images, so should not affect the other Tornado work

samples/jigsaw-prototype.py
```diff
@@ -0,0 +1,18 @@
...
+pixelYscale = int(img.shape[1] / cols)
+
+for r in range(1, rows + 1):
+    for c in range(1, cols + 1):
+        cv2.imwrite(
+            f"images/img{r}_{c}.png",
+            img[
+                (r - 1) * pixelXscale : r * pixelXscale,  # noqa: E203
+                (c - 1) * pixelYscale : c * pixelYscale,  # noqa: E203
+                :,
```
aaronabraham311, line 16: Weird question but what's the colon for?

aaronabraham311 (approved):
LGTM. Have we tested it on skewed images, like extremely tall or wide images?

samples/images/.gitignore
ahmedhamodi: Multiple .gitignore files might not be the way to go, but it is how I've done it in the past.

It will only ignore files in it's immediate directory, not parent directories.

https://riptutorial.com/git/example/2888/ignoring-files-in-subfolders--multiple-gitignore-files-

ahmedhamodi:
> LGTM. Have we tested it on skewed images, like extremely tall or wide images?

Since rows and columns are calculated differently, it will still create 9 images. But in general, let's avoid extremely skewed images for our rooms.

samples/jigsaw-prototype.py
ahmedhamodi: Actually not entirely sure, it was in the example I was following. However, I did deviate from the example with how I am doing the image pixel calculations, and I just tested without that third parameter and it works.

Will just remove.

aaronabraham311 (approved):

jaydhulia (approved):

PR #56, opened 2021-03-27
Aframe - basic scene, object, background setup

jaydhulia:
- [x] Populate scene data from backend
- [x] Populate object data from backend
- [x] Render asset-items based on response from backend
- [x] Render objects based on response from backend
- [x] Render background based on response from backend
- [x] Render camera properties based on response from backend
- [x] Click animations work


ahmedhamodi (commented):
What are these .glb files? I don't think we should be committing them to the github repo.

jaydhulia:
@ahmedhamodi The .glb files are just the compressed version of the .gltf that existed in the `assets` folder already. It is okay to leave just these ones in here, no more will be added.
The files that designers choose will go in the Google drive. The files in this PR are small enough to not cause issues.

ahmedhamodi (approved):

PR #57, opened 2021-03-29
Popup function and html structure.

drbalar:
switches between 2 cameras to simulate popups. It is a supported thing by the [camera component](https://github.com/aframevr/aframe/blob/master/docs/components/camera.md) and clean too.

`addEntityToBlackboard()`  is the function that populates the entities onto the blackboard. Currently only 1 supported. Will add support for creating multiple elements


samples/chair.js
```diff
@@ -0,0 +1,33 @@
+AFRAME.registerComponent("chair", {
```
ahmedhamodi, line 1: Is chair still an appropriate name for this?

samples/popup.html
```diff
@@ -0,0 +1,204 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Popup Demo</title>
+    <meta name="description" content="Popup demo" />
+    <script src="https://aframe.io/releases/1.2.0/aframe.min.js"></script>
+    <script src="chair.js"></script>
+    <script>
+      // TODO: Verify room is not needed and delete it
```
jaydhulia, line 10: Is this TODO part of this PR? With gravity gone, we may not need an actual room 😄

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+      AFRAME.registerComponent("read-room", {
```
jaydhulia, line 11: Can we move the component registration in a diff file? In the actual repo structure all of these will go in `public/static/js/components/*`

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+
+          var eastWall = document.createElement("a-box");
+          eastWall.setAttribute("width", "20");
+          eastWall.setAttribute("height", "20");
+          eastWall.setAttribute("depth", "1");
+          eastWall.setAttribute("rotation", "0 -90 0");
+          eastWall.setAttribute("position", "10 0 0");
+          eastWall.setAttribute("color", "yellow");
+          eastWall.setAttribute("opacity", "0.25");
+          eastWall.setAttribute("static-body", "");
```
jaydhulia, line 93: We can get rid of the static-body, kinematic-body etc stuff, as we won't be using gravity :)

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+          el.appendChild(eastWall);
+        },
+      });
+    </script>
+  </head>
+  <body>
+    <a-scene
+      background="color: #ECECEC"
+      cursor="rayOrigin: mouse"
+      physics="debug: true; gravity: 0"
```
jaydhulia, line 103: SImilarly, we can completely get rid of `physics`

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+  var jsonDataArr = [
+    'jsonData: {"width": "1", "height": "2", "depth": "1", "color": "orange", "x": "2", "y": "0", "z": "0"}',
+    'jsonData: {"width": "1", "height": "2", "depth": "1", "color": "blue", "x": "0", "y": "0", "z": "0"}',
+  ];
+
+  setupEnv(jsonDataArr);
+
+  function setupEnv(jsonDataArr) {
+    // default return button. Always here.
+    // TODO: move into a A-Frame component. [closePopup]
```
jaydhulia, line 158: This is a good idea; we should move this to a component

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+  function addEntityToBlackboard(componentData) {
+    var componentDataParsed = JSON.parse(componentData);
+    var blackboardEl = document.querySelector("#blackboard");
+    var entityEl = document.createElement("a-entity");
+    entityEl.setAttribute(
+      componentDataParsed.componentType,
+      "jsonData",
+      JSON.stringify(componentDataParsed.jsonData)
+    );
+    blackboardEl.appendChild(entityEl);
```
jaydhulia, line 195: I think `appendChild` is an async function. So before switching the camera, I believe we should listen for the `loaded` event: https://aframe.io/docs/1.2.0/introduction/javascript-events-dom-apis.html#adding-an-entity-with-appendchild

samples/popup.html
```diff
@@ -0,0 +1,204 @@
...
+        animation__rotationcompletecomplete="property: rotation; to: 0 360 0; dur:500; startEvents: animationcomplete__positioncomplete"
+      ></a-box>
+      <a-entity
+        id="primaryCamera"
+        position="0 0 9"
+        rotation="0 0 0"
+        camera
+        look-controls
+        wasd-controls
+        kinematic-body
```
jaydhulia, line 126: We should be able to get rid of this `kinematic-body`

samples/popup.html
drbalar: yes, so i can remove that now i guess. I just wanted to confirm that

samples/chair.js
drbalar: not at all lol. It was the name we used in all the other samples so far so i didnt change it. I will change now

samples/popup/popup.html
```diff
@@ -0,0 +1,96 @@
...
+        position="0 0 9"
+        rotation="0 0 0"
+        camera
+        look-controls
+        wasd-controls
+      ></a-entity>
+      <a-entity id="readRoom" position="0 -100 0">
+        <a-entity
+          id="blackboard"
+          info-panel
```
jaydhulia, line 23: Is `info-panel` an Aframe thingy? Or just leftover from something?

drbalar: leftover, will remove

samples/popup.html
drbalar: I added a listener although I don't think it is having any impact on it in our case. I might be mistaken.

samples/popup/popup.html
```diff
@@ -0,0 +1,62 @@
...
+      ></a-entity>
+      <a-entity id="readRoom" position="0 -100 0">
+        <a-entity
+          id="blackboard"
+          visible="true"
+          geometry="primitive: plane; width: 18   ; height: 15"
+          material="color: #333333; shader: flat; transparent: false"
+        >
+        </a-entity>
+        <a-entity id="returnButton"></a-entity>
```
jaydhulia, line 30: I believe instead of initializing this in JS, (like on lines 47-51), you should be able to do this directly in here, like:
```
<a-entity id="returnButton" close-popup="json_data: <json_data>"></a-entity>
```

samples/popup/components/openPopup.js
```diff
@@ -0,0 +1,56 @@
...
+    // Create material.
+    this.material = new THREE.MeshStandardMaterial({ color: data.color });
+
+    // Create mesh.
+    this.mesh = new THREE.Mesh(this.geometry, this.material);
+
+    el.object3D.position.set(data.x, data.y, data.z);
+    // Set mesh on entity.
+    el.setObject3D("mesh", this.mesh);
+    el.addEventListener("loaded", function (e) {
```
jaydhulia, line 32: Sorry, I should have been clearer, I meant we need to listen for the `loaded` on line 55, as `blackboardEl.appendChild(entityEl);` is the call that is async. So, we need to do this in the opposite order.
Like:
```
el.addEventListener("click", function () {
          entityEl = addEntityToBlackboard(JSON.stringify(data.blackboardData));
          // add a listener to entityEl being `loaded` and do lines 37 and 38 in there. 
```
Does that make sense? Let me know if that's confusing

jaydhulia: That would mean `entityEl` needs to be returned from line 55. Or you could do `// add a listener to entityEl being `loaded` and do lines 37 and 38 in there` inside of there

samples/popup.html
jaydhulia: Yeah, for now since we have basic rendering it won't affect it much, but it will affect it if we have more complex objects being rendered :) Like puzzles

samples/popup/components/openPopup.js
jaydhulia: PS, check slack :)

jaydhulia (approved):
Woho, looks great! Thanks @drbalar

PR #59, opened 2021-04-02
Integrate Popup + Keypad

jaydhulia:
- [X] Add a generic popup to Aframe
- [X] Retrieve component to display in popup from backend
- [X] Update mockdata to account for changes
- [X] Integrate a Keypad along with required logic + passcode from backend
- [x] Add a label for the Keypad


app/public/static/js/components/keypad.js
```diff
@@ -0,0 +1,49 @@
...
+    el.setAttribute("position", { x: 0, y: 0, z: 0.5 });
+
+    el.setAttribute("class", "link");
+
+    el.addEventListener("superkeyboardinput", function (event) {
+      if (event.detail.value === password) {
+        el.setAttribute(
+          "super-keyboard",
+          "label:SUCCESS; labelColor: green; multipleInputs:true"
+        );
```
ahmedhamodi, line 30: Personally I think auto closing (after 1.5 seconds or so) on a success makes sense. But I'll leave the decision to you guys

ahmedhamodi (commented):
One comment. Otherwise code looks good to me.

Will (attempt to) run locally now. If it works, I'll approve the PR.

ahmedhamodi (approved):
Works locally :)

app/public/static/js/components/keypad.js
jaydhulia: Synced offline about this; the success workflow will be worked on later once we have state

PR #61, opened 2021-04-03
Add instructions to run tornado backend on vscode IDE to readme.

ahmedhamodi:
Will also add the following items to the readme:
- "port 5432 already in use" error resolution
- <todo>


ahmedhamodi:
We may or may not need to land this ever, but it's mostly for me so I don't need to ask for help on the command to run the tornado backend on VS Code :)

PR #62, opened 2021-04-03
Update instructions for running backend server

jaydhulia:


README.md
```diff
@@ -45,19 +45,41 @@ Note: Every time you run `git commit` in this repo, it will run some lint checks
...
+1. Run `export PYTHONPATH=.`
+1. Run `export CONFIG_PATH=configs/dev-config.yaml`
+1. If you have not done so recently, run `make install` as it will update any schema changes.
+1. Run `python app/__main__.py`
+1. You should see a `SERVER STARTED` message along with configuration details
+
+#### Windows
+
+1. Ensure you are in the root directory, and your virtual env is activated. Also make sure your Postgres is running.
+1. Run `set PYTHONPATH=.`
```
drbalar, line 74: If they using bash on windows, it will be `export PYTHONPATH=.` and `export CONFIG_PATH=configs/dev-config.yaml` too, but looks good

drbalar (dismissed):

README.md
jaydhulia: Does `set` still work if they are using bash on Windows? Or does it have to be `export` strictly?

drbalar: It depends on the terminal they use. For `git bash` only `export` works. For `cmd` i think it is `set`

jaydhulia: Updated instructions; Windows why are you like this 😞

drbalar (approved):
thank you, windows side looks good. Mac looks good too, but i could not test.

README.md
```diff
@@ -1,16 +1,16 @@
...
 
 ---
 
-## About Calgary Distress Center
+# About Calgary Distress Center
 
 todo
 
-### Team
+# Team
```
ahmedhamodi, line 13: Any reason for making Team/about/getting started all on the same level (and using the bigger header)? It's just inconsistent with frontend now. Not a big deal I guess

jaydhulia: The .md format only seems to support nesting to 4 `####` and for running the backend repo topic, I wanted to subdivide the sections. But with the current file, I wasn't able to do that. Then I noticed that we started at 2 `##` rather than 1 `#`, so I moved everything up 1 header.

ahmedhamodi: Makes sense. The contrast from `#` to paragraph text just looks so bad for me 😭  but it's okay

PR #63, opened 2021-04-06
Aframe Text Pane

vivianliu0:
Added text pane component that displays a list of text and allows users to go back and forth between text using navigation buttons.

Link to loom video: https://www.loom.com/share/72126733a9cc4b5e967227e55930dc86
Password: text-pane

Screenshot of text pane:
image


samples/text_panes/text_pane.html
```diff
@@ -0,0 +1,35 @@
...
+        <a-entity
+          id="blackboard"
+          visible="true"
+          geometry="primitive: plane; width: 18; height: 15"
+          material="color: #333333; shader: flat; transparent: false"
+        ></a-entity>
+        <a-entity
+          id="text_pane"
+          text_pane='jsonData: {"text": ["Welcome to the Escape Room! This is a very loooooong sentence for testing. This is a test this is a test this is a test this is a test.","This is where text will go. This is where text will go. This is where text will go", "This is some more text. This is some more text. This is some more text."], "currPosition" : 0}'
+        >
```
ahmedhamodi, line 29: @jaydhulia I forget how we did other stuff, but is there one more level of extraction we want to use here or is this fine for now?

samples/text_panes/components/prism.js
```diff
@@ -0,0 +1,33 @@
+AFRAME.registerComponent("prism", {
```
ahmedhamodi, line 1: Not entirely sure where this is being use, can we use a more descriptive name?

ahmedhamodi (commented):
Overall lgtm but I'll let Jay make the final approval since he has more aframe context than I do at this point. Couple small comments.

Also, would love to see a loom video as you have identified in your todo! :)

samples/text_panes/components/text_pane.js
```diff
@@ -0,0 +1,97 @@
...
+    this.textLabel.setAttribute("position", "-4.9 0 0.25");
+    this.el.appendChild(this.textLabel);
+
+    if (data.text.length > 1) {
+      const leftNavProp =
+        '{"width": "1.5", "height": "1.5", "depth": "0.01", "color": "white", "x": "-8", "y": "0", "z": "0.005"}';
+      const rightNavProp =
+        '{"width": "1.5", "height": "1.5", "depth": "0.01", "color": "white", "x": "8", "y": "0", "z": "0.005"}';
+
+      // Create left nav textLabel
```
jaydhulia, line 30: For the sake of optimization, let's combine the nav textlabel, and the nav button into one entity. I think aframe has that feature which may be useful. Like:
```
<a-entity
  geometry="primitive: plane; width: 4; height: auto"
  material="color: blue"
  text="value: This text will be 4 units wide."></a-entity>
```
Basically something like the above, but in JS. This means that your `prism` component can be renamed to like `text-box`, and additionally take in some more fields related to the text and its properties and set it appropriately. 
So, instead of needing to keep the box and text in sync, we create an entity that combines the two of them.

samples/text_panes/components/text_pane.js
```diff
@@ -0,0 +1,97 @@
...
+      this.buttonLabelLeft.setAttribute("scale", "2 2 1");
+      this.buttonLabelLeft.setAttribute("color", "black");
+      this.buttonLabelLeft.setAttribute("position", "-8.5 0 0.01");
+      this.el.appendChild(this.buttonLabelLeft);
+
+      // Create left nav button
+      this.leftNav = document.createElement("a-entity");
+      this.leftNav.setAttribute("id", "nav-button-left");
+      this.leftNav.setAttribute("prism", "jsonData", leftNavProp);
+      this.leftNav.addEventListener("click", function () {
```
jaydhulia, line 44: Ran into a small bug, where clicking on the text directly doesn't cause the click to be registered (like clicking on `Next`, but clicking on the white part does. I believe this is because the text is slightly in front of the button, and only the button has a click handler. This bug should be solved by my previous comment though.

samples/text_panes/components/text_pane.js
```diff
@@ -0,0 +1,97 @@
...
+        if (data.currPosition != 0) {
+          --data.currPosition;
+          document
+            .querySelector("#text")
+            .setAttribute("value", data.text[data.currPosition]);
+          document
```
jaydhulia, line 50: Is there a reason for this part here? The part about `.setAttribute("jsonData", rightNavProp);` for the `#text-nav-right`?

jaydhulia: I think we want to reset the text to be `Next` here right? Right now going all the way to the end, it is `Done`, but doesn't reset to `Next` on going left

samples/text_panes/components/text_pane.js
```diff
@@ -0,0 +1,97 @@
...
+      this.el.appendChild(this.buttonLabelLeft);
+
+      // Create left nav button
+      this.leftNav = document.createElement("a-entity");
+      this.leftNav.setAttribute("id", "nav-button-left");
+      this.leftNav.setAttribute("prism", "jsonData", leftNavProp);
+      this.leftNav.addEventListener("click", function () {
+        if (data.currPosition != 0) {
+          --data.currPosition;
+          document
```
jaydhulia, line 47: For these query selector statements, I believe we should be able to skip those and use references directly 😄 . Like, above the addEventListener:
```
const textLabelConst = this.textLabel;
```
and then inside it:
```
textLabelConst.setAttribute("value", data.text[data.currPosition]);
```

samples/text_panes/text_pane.html
jaydhulia: This should be good. This is what it is going to be stored in Postgres and will be passed to the component `text_pane` via the Blackboard

samples/text_panes/components/prism.js
vivianliu0: My bad, the naming was ambiguous, changed it to text-box

samples/text_panes/components/text_pane.js
vivianliu0: Combined the textlabel and button into one text-box entity, lmk if this is what you had in mind

samples/text_panes/components/text_pane.js
vivianliu0: Thanks for catching that! The bug should be resolved now

samples/text_panes/components/text_pane.js
vivianliu0: Oops yea should be reset to next when the left nav button is clicked, should be fixed now

samples/text_panes/components/text_pane.js
vivianliu0: TYYY oh my this was giving me problems before but it's working noww :))

jaydhulia (approved):
Beautiful, this is exactly what I meant! Looks great to me 😄

ahmedhamodi:
Thanks for the loom video :) Good to merge!

PR #64, opened 2021-04-06
Add server-side ui caching

jaydhulia:
- [x] Added check for server-side caching for scenario
- [x] Added check for server-side caching for scene
- [x] Added a new cache layer for scenario friendly name <> scenario_id cache

### DIFF
--- a/app/routes/ui/scenario.py
+++ b/app/routes/ui/scenario.py
@@ -61,7 +61,6 @@ class UIScenarioHandler(BaseUIHandler):
             )
         except ValueError as e:
             self.write_error(status_code=404, message=str(e))
-        except Exception as e:
+        except Exception:
             # no need to expose exact exception error on this route
-            print(str(e))
             self.write_error(status_code=500, message="Internal Server Error")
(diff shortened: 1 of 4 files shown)


app/routes/ui/scenario.py
```diff
@@ -17,20 +19,35 @@ class UIScenarioHandler(BaseUIHandler):
...
             scene_number_int = int(scene_number)
-            # TODO: caching will be dealt with in a later PR
-            scenario_obj: Scenario = await get_scenario_by_friendly_name_from_postgres(
-                scenario_friendly_name, self.db_session
-            )
+            scenario_dict = await check_and_get_scenario_by_name(scenario_friendly_name)
+            if not scenario_dict:
+                if not self.db_session:
+                    self.db_session = get_session()
+                scenario_obj: Scenario = (
```
ahmedhamodi, line 26: Never seen this syntax before! Mind explaining quickly what is happening here?

ahmedhamodi (approved):
lgtm! Had one clarifying question for my own knowledge, but I trust that it works as expected :)

app/routes/ui/scenario.py
jaydhulia: Defining `scenario_obj` as type `Scenario`, and the bracket is just something that was added by the linters to make it prettier, to keep the function call on one line!

PR #65, opened 2021-04-06
For updating cache periodically

jaydhulia:
- [x] Update cache on a timer based on config settings
- [x] Enable/Disable enabling of caching
- [x] Update Scenario cache
- [x] Update Scene cache
- [x] Update asset cache

### DIFF
--- a/app/cache/update_cache.py
+++ b/app/cache/update_cache.py
@@ -1,13 +1,28 @@
+import time
+from datetime import datetime
+
+from cache.cache import clear_all_cache, update_scenario_cache
+from library.postgres import get_scene_from_postgres
 from models import get_session
 from models.db_client import get_scenarios
 
 
 async def update_cache():
     session = get_session()
+    clear_all_cache()
     scenarios = get_scenarios(session)
+    scenes_updated = set()
     for scenario in scenarios:
-        # TODO: include scenario's friendly_name from other PR in here
-        # await update_scenario_cache(scenario.id, scenario.as_dict())
-        pass
-    # update scene cache
-    # update asset cache
+        await update_scenario_cache(
+            scenario.id, scenario.as_dict(), scenario.friendly_name
+        )
+        for scene_id in scenario.scene_ids:
+            if scene_id not in scenes_updated:
+                await get_scene_from_postgres(
+                    scene_id,
+                    session,
+                    update_cache=True,
+                )
+                scenes_updated.add(scene_id)
+
+    print(datetime.fromtimestamp(int(time.time())), "Cache updated")
--- a/app/cache/update_cache.py
+++ b/app/cache/update_cache.py
@@ -9,7 +9,7 @@ from models.db_client import get_scenarios
 
 async def update_cache():
     session = get_session()
-    clear_all_cache()
+    await clear_all_cache()
     scenarios = get_scenarios(session)
     scenes_updated = set()
     for scenario in scenarios:
--- a/app/cache/update_cache.py
+++ b/app/cache/update_cache.py
@@ -1,0 +1,13 @@
+from models import get_session
+from models.db_client import get_scenarios
+
+
+async def update_cache():
+    session = get_session()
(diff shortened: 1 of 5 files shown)


app/library/postgres.py
```diff
@@ -34,15 +34,15 @@ async def post_asset_to_postgres(data: dict, session):
     return asset_model.as_dict()
 
 
-async def get_object_from_postgres(object_id: str, session):
+async def get_object_from_postgres(object_id: str, session, update_cache=False):
```
drbalar, line 37: `update_cache` is always `True` regardless of the config value.. I'm guessing that is going to be implemented in another PR?

ahmedhamodi: Did you mean always false? and I think this is a default value for if the field isn't set

drbalar: The function that calls `get_object_from_postgres` is always passing in true

ahmedhamodi: Ah I see, lines 21-25 in update_cache.py. Good point

configs/dev-config.yaml
```diff
@@ -30,3 +30,7 @@ script:
 cache:
   # seconds in day: 24 * 60 * 60
   expire_time: 86400
+  # how frequently to update cache (in seconds), 60 seconds for dev, for now
+  update_time: 5
```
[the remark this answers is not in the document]
jaydhulia: Yup, this should be 60 if we actually enable it, but since we aren't enabling it for dev, it doesn't matter 😄

app/cache/update_cache.py
```diff
@@ -0,0 +1,28 @@
...
+
+from cache.cache import clear_all_cache, update_scenario_cache
+from library.postgres import get_scene_from_postgres
+from models import get_session
+from models.db_client import get_scenarios
+
+
+async def update_cache():
+    session = get_session()
+    clear_all_cache()
```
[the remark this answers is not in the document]
jaydhulia: Ah good catch, will update

app/library/postgres.py
jaydhulia: Nope this is meant to be always true in that function.
The config's `update_enabled` being `True` or `False` determines whether this function runs or not. 
If this function runs, the `update_cache` parameter should always be true, otherwise there is no point in running this function 😄 Does that make sense?

app/cache/update_cache.py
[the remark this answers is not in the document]
jaydhulia: Updated!

app/library/postgres.py
jaydhulia: An example of where `update_cache` would be `False` is the admin routes. For the user routes, and the update_cache.py we always want it to be true, so that it updates the cache. The logic for the config's `update_enabled` being `True` or `False` is handled in main (https://github.com/uwblueprint/cdc-backend/pull/65/files#diff-a4bc3d28ebe484212cebe359972290c5dd689f559c1dda8d6f71014815f313d1R109)

drbalar: Ahhh got it. So the config's `update_enabled` only affects the admin routes, not the user routes. That makes sense, ty

drbalar (approved):
LG

app/library/postgres.py
jaydhulia: Well no, sorry, to clarify:
- config's `update_enabled` determines whether calling the `update_cache` method periodically is enabled or not.
- the parameter `update_cache` is always False for admin routes, and always True for user routes, independent of config's `update_enabled`

Does that make sense?

drbalar: Oh got it now. The naming tripped me lol

jaydhulia: Fair enough, I will update the config's name to `periodic_callback_enabled`

jaydhulia: Updated, please re-approve when you get a chance, ty!

drbalar (approved):
thank you :D

PR #66, opened 2021-04-09
Add Nav mesh and constrain movement based on nav-mesh

jaydhulia:
Notes:
- nav-mesh doesn't have support for aframe v1.2.0 (not really sure when / if it will be added)
- Downgrading to aframe v1.1.0 hasn't broken anything that we have so far
- nav-mesh works with .gltf files as meshes but not blender generated .glb files
- nav-mesh doesn't work if `wasd-controls` are on the camera


ahmedhamodi (commented):
Code looks good. Going to test locally at start of work session

app/routes/ui/scenario.py
```diff
@@ -50,13 +50,22 @@ async def get(self, scenario_friendly_name, scene_number=0):
                 )
 
             is_last_scene = scene_number_int == len(scenario_obj.scene_ids) - 1
+            background_ext = (
+                "." + scene_dict["background_details"]["s3_key"].split(".")[-1]
```
drbalar, line 54: Does [-1] mean the last element in the array? Or a typo?

app/public/templates/scene_nav_mesh.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
drbalar, line 1: Where is this one used? Is it a sample?

app/routes/ui/scenario.py
ahmedhamodi: Yeah last element in array!

app/public/templates/scene_nav_mesh.html
jaydhulia: Yeah it is for creating the nav-mesh for a scene. For example, if we have a new scene, just change the render template here https://github.com/uwblueprint/cdc-backend/blob/master/app/routes/ui/scenario.py#L54 to "scene_nav_mesh.html" 😄 This will be useful for creating the nav-mesh objects once we have the actual scenes from designers

app/routes/ui/scenario.py
jaydhulia: Yeah basically trying to get the extension of the file, so would be splitting by "." and getting last element

ahmedhamodi (approved):
Works great!

app/public/templates/scene_nav_mesh.html
drbalar: Got it, great!

drbalar (approved):
Looks good tmtooo!

PR #68, opened 2021-04-10
Update readme to resolve port in use error

ahmedhamodi:
An issue I've run into like 30 times, and even today when booting up Jay's navigation mesh change. Thought it would be good to include in the readme in case other people run into this.


README.md
```diff
@@ -68,6 +68,8 @@ Once you have the requirements installed, you should be able to develop by just
 1. Run `python app/__main__.py`
 1. You should see a `SERVER STARTED` message along with configuration details
 
+Note, if you get a "port 5432 in use" on the postgres app, run `sudo pkill -u postgres` to clear the port and then restart postgres.
```
jaydhulia, line 71: Uhmm instead of doing this, let's do it this way:
1. Find the pid of the process using that port: `sudo lsof -n -i :5432`
2. Kill the process by using the PID from above result: `sudo kill -9 <PID>`

jaydhulia: (because it may be other processes using that port)

jaydhulia (approved):

PR #69, opened 2021-04-10
Integrate text-panes with blackboard

jaydhulia:
Text panes work with blackboard now 🎉 . Click on the left most chair to see the magic

### DIFF
--- a/app/public/templates/scene.html
+++ b/app/public/templates/scene.html
@@ -9,9 +9,11 @@
     <script src="https://unpkg.com/aframe-super-keyboard@2.1.0/dist/aframe-super-keyboard.min.js"></script>
 
     <!-- A-frame components   -->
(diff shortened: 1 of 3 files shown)


app/public/templates/scene.html
```diff
@@ -9,9 +9,11 @@
     <script src="https://unpkg.com/aframe-super-keyboard@2.1.0/dist/aframe-super-keyboard.min.js"></script>
 
     <!-- A-frame components   -->
+    <!--  TODO: minify and combine all components once we are done  -->
```
ahmedhamodi, line 12: 100%

ahmedhamodi (commented):
Code looks good. Will test locally

ahmedhamodi (approved):
🎉 works really well.

PR #71, opened 2021-04-11
Aframe Ordered Puzzles

vivianliu0:
What's completed in this PR:
- Given a list of images with an initial position and target location, these images are all populated onto a blackboard (Note that the number of images provided by admin can be arbitrary)
- Users can then drag images to different target locations. If all pictures are correctly placed in their target location, the blackboard is cleared

Here's a quick demo of the puzzle:
https://www.loom.com/share/f7e0c74942694ce398f421781e07b507


ahmedhamodi:
The loom demo is so cool, thank you so much for doing this!

Also tagging @aaronabraham311 to get product spec approval on this PR.

ahmedhamodi:
My thought to support the jigsaw puzzle prototype is that instead of a 1xn array of final positions, we make it a nxm matrix, allowing for a 3x3 jigsaw puzzle solution.

@vivianliu0 - there is a separate ticket for this extension and I will assign it to you shortly. So you can work on this in a separate PR to avoid this one getting too large.

@aaronabraham311  - thoughts on this as a final product / design flow?

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    Math.abs(originalPos["x"] + delta["x"] - targetPos["x"]) <= accuracyReq &&
+    Math.abs(originalPos["y"] + delta["y"] - targetPos["y"]) <= accuracyReq
+  );
+}
+
+function isPuzzleComplete() {
+  var puzzlePieces = document.querySelectorAll("a-image.draggable");
+  //var puzzlePieces = document.querySelector("#ordered-puzzle").children;
+  console.log(puzzlePieces);
```
ahmedhamodi, line 79: Let's remove this console log

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function isOnTarget(puzzlePiece) {
+  const delta = puzzlePiece.object3D.children[0].position;
+  const originalPos = puzzlePiece.getAttribute("position");
+  const targetPos = {
+    x: puzzlePiece.getAttribute("xTarget"),
+    y: puzzlePiece.getAttribute("yTarget"),
+  };
+  const accuracyReq = 1;
```
ahmedhamodi, line 69: Did you try playing around with this value? Is 1 a good amount to leave some tolerance for the user?

ahmedhamodi (commented):
Code looks good to me. Will differ to Dhruvin and Jay for more feedback on the quality of code given they have more Aframe experience at this point.

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+  multiple: true,
+
+  init: function () {
+    const data = this.data.jsonData;
+    const el = this.el;
+
+    const numPuzzlePieces = data.images.length;
+
+    for (i = 0; i < numPuzzlePieces; i++) {
+      var textBoxProp =
```
jaydhulia, line 18: Instead of manually creating the string version of the JSON object here, you should just be able to do:
```
JSON.stringify(data.images[i])
```
and pass it directly like:
```
this.target.setAttribute("target-box", "jsonData", JSON.stringify(data.images[i]));
```
Note, if you want to add something like color, you will have to do it before stringifying it, like
```
data.images[i].color = "yellow"
```

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+        '{"width": "' +
+        data.images[i].width +
+        '", "height": "' +
+        data.images[i].height +
+        '", "color": "yellow", "x": "' +
+        data.images[i].xTarget +
+        '", "y": "' +
+        data.images[i].yTarget +
+        '", "z": "0", "text": "' +
```
jaydhulia, line 27: Is there a reason for text: i here? Is that being used in target-box?

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+        i +
+        '"}';
+
+      this.target = document.createElement("a-entity");
+      this.target.setAttribute("id", "puzzle-target");
```
jaydhulia, line 32: This will cause duplicate id for all the puzzle pieces, so let's add an extra identifier, like "puzzle-target-" + i

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+      this.puzzlePiece.setAttribute("width", data.images[i].width);
+      this.puzzlePiece.setAttribute("height", data.images[i].height);
+      this.puzzlePiece.setAttribute(
+        "position",
+        data.images[i].x + " " + data.images[i].y + " 0"
+      );
+      this.puzzlePiece.setAttribute("xTarget", data.images[i].xTarget);
+      this.puzzlePiece.setAttribute("yTarget", data.images[i].yTarget);
+      this.puzzlePiece.setAttribute("onTarget", false);
+      this.puzzlePiece.setAttribute("class", "draggable");
```
jaydhulia, line 48: We need to add class link here as well for compatibility with the blackboard stuff

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+      parse: JSON.parse,
+      stringify: JSON.stringify,
+    },
+  },
+
+  multiple: true,
+
+  init: function () {
+    const data = this.data.jsonData;
+    const el = this.el;
```
jaydhulia, line 13: Since you are using this.el below, const el = this.el can be removed 😄

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+function onDragEnd(event) {
+  event.target.setAttribute("onTarget", isOnTarget(event.target));
+  if (isPuzzleComplete()) {
+    closePuzzle();
+  }
+}
+
+function isOnTarget(puzzlePiece) {
+  const delta = puzzlePiece.object3D.children[0].position;
+  const originalPos = puzzlePiece.getAttribute("position");
```
jaydhulia, line 64: I am a little bit confused here... Why are the two positions different? Like puzzlePiece.object3D.children[0].position vs puzzlePiece.getAttribute("position"), do you mind explaining what each of them represent?

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+  };
+  const accuracyReq = 1;
+  return (
+    Math.abs(originalPos["x"] + delta["x"] - targetPos["x"]) <= accuracyReq &&
+    Math.abs(originalPos["y"] + delta["y"] - targetPos["y"]) <= accuracyReq
+  );
+}
+
+function isPuzzleComplete() {
+  var puzzlePieces = document.querySelectorAll("a-image.draggable");
```
jaydhulia, line 77: Unfortunately, document.querySelectorAll can really affect performance, especially if it's being called often. Instead, let's add a cached list of all the puzzle pieces.
One way to do this is, in your for loop above where you do `this.el.appendChild(this.puzzlePiece);`, also add it to a cached list like `this.cachedPuzzlePieces.append(this.puzzlePiece)`.
Then in a second for loop right after, add the event handlers `onDragEnd`, and pass the cachedPuzzlePiecesConst to it as well. Then you should be able to iterate through cachedPuzzlePiecesConst like you currently do.

Note: I haven't actually tried this, so you may have to play around with my suggestion to get it working!

samples/ordered_puzzle/components/target_box.js
```diff
@@ -0,0 +1,29 @@
...
+    // Create geometry.
+    this.geometry = new THREE.PlaneGeometry(data.width, data.height);
+
+    // Create material.
+    this.material = new THREE.MeshStandardMaterial({ color: data.color });
+
+    // Create mesh.
+    this.mesh = new THREE.Mesh(this.geometry, this.material);
+
+    el.object3D.position.set(data.x, data.y, data.z);
```
jaydhulia, line 25: In accordance with above comments, this will have to change to like `data.xTarget` instead of `data.x` etc

samples/ordered_puzzle/ordered_puzzle.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="https://aframe.io/releases/1.1.0/aframe.min.js"></script>
+    <script src="https://unpkg.com/aframe-drag-controls@1.0.0/index.js"></script>
+    <script src="components/ordered_puzzle.js"></script>
+    <script src="components/target_box.js"></script>
+  </head>
+  <body>
+    <a-scene background="color: #AAD5D7" cursor="rayOrigin: mouse">
```
jaydhulia, line 10: For compatibility reasons, you will want to add raycaster here, like:
```
<a-scene background="color: #AAD5D7"cursor="rayOrigin:mouse" raycaster="objects: .link">
```

samples/ordered_puzzle/components/ordered_puzzle.js
```diff
@@ -0,0 +1,94 @@
...
+        i +
+        '"}';
+
+      this.target = document.createElement("a-entity");
+      this.target.setAttribute("id", "puzzle-target");
+      this.target.setAttribute("target-box", "jsonData", textBoxProp);
+      this.el.appendChild(this.target);
+
+      this.puzzlePiece = document.createElement("a-image");
+      this.puzzlePiece.setAttribute("id", "puzzle-piece-image");
```
jaydhulia, line 37: This will cause duplicate id for all the puzzle pieces, so let's add an extra identifier, like "puzzle-piece-image-" + i

jaydhulia (commented):
Works great, nice work 🎉 ! Added some minor comments for code changes to make it prod ready, but otherwise looks great!

samples/ordered_puzzle/ordered_puzzle.html
```diff
@@ -0,0 +1,39 @@
...
+        <a-entity
+          id="blackboard"
+          visible="true"
+          geometry="primitive: plane; width: 22; height: 15"
+          material="color: #333333; shader: flat; transparent: false"
+        ></a-entity>
+        <a-entity
+          id="ordered-puzzle"
+          ordered-puzzle='jsonData: {"images": [{"imageSrc" : "./images/pic1.png", "width": 3, "height": 3, "xTarget": -7, "yTarget": 0, "x": 0, "y": 4},
+          {"imageSrc" : "./images/pic2.png", "width": 3, "height": 3, "xTarget": -3.5, "yTarget": 0, "x": -2, "y": -4},
```
jaydhulia, line 30: @aaronabraham311 @ahmedhamodi What are your thoughts here? I am wondering whether it's feasible (technically and logically) for admins to define the target coordinates as well, or we should get what the right order should be, and just define our xTarget, yTarget to be something static on frontend side?
Like as an example, the top row would be the randomized order images, and then below that would always be the static row of "Targets", and the number of targets to show would be = num pieces.

ahmedhamodi: This is a good point. I'm not sure how to go about getting the target locations dynamically without making it look off center or weird. Should we just support 3-4 different layouts and have those (for lack of better term) hardcoded? Something like 1x5, 3x3, and 2x2 would be the 3 I think we should support. Perhaps even 1x4.

Then I agree that having the images load at the top of the popup, in any format/location, works well and having the target locations fixed based on the layout configuration. Open to thoughts and suggestions here though.

samples/ordered_puzzle/components/ordered_puzzle.js
vivianliu0: It's currently not being used in target-box, but I was thinking in the future if we wanted to include text on the target boxes that could also be an option. I'll remove it in the PR since it's not being used.

samples/ordered_puzzle/components/ordered_puzzle.js
vivianliu0: I originally thought that these two positions were the same as well, but when I was testing it out it turns out the puzzlePiece.object3D.children[0].position represents the position of the object relative to the initial position that was set by puzzlePiece.getAttribute("position").

samples/ordered_puzzle/components/ordered_puzzle.js
vivianliu0: I was playing around with this value and currently the tolerance of 1 means that as long as the majority of the puzzle piece is on the target, it is considered to be on target.

samples/ordered_puzzle/ordered_puzzle.html
aaronabraham311: There is a potentially simpler fix for this issue. We just don't have targets. 

In the ticket, I mentioned that the clues should reveal themselves. We really don't need to check order:

> - The user will need to order the 5 pictures in a way that gives a clue
>    - For example: the clue is mouse, the password to get out of the room. The images will be the letters of mouse and once arranged properly, will reveal the clue

Maybe another example would help clarify this. Let's say that one of the clues for a puzzle is something to do with the word 'teapot'. We would have all the pictures for each letter jumbled up, but as the user shuffles it around, they realize that it's the word 'teapot'. There is no need to define targets in this case, because the clue reveals itself when you shuffle it around. This makes things simple technically and we can have a variety of use cases for the admin (jigsaw, word plays, visual order like Vivian's example, etc.). As a general product principle: if we want users to be creative, remove as much structure as possible so they have freedom to play around. 

For now, what Vivian created is great and potentially would help with jigsaw puzzles, but I am seriously thinking whether it is possible to remove the order checking and just have the blackboard with all the images. We can use her order checking for a puzzle like jigsaw.

aaronabraham311:
> Also tagging @aaronabraham311 to get product spec approval on this PR.

First of all, excellent work Vivian! I wasn't expecting you to implement order checking (wasn't in the ticket) but it looks awesome.

As mentioned in a previous comment, this is a little hard to generalize for admins. I propose that we generalize this A-frame component where we have order checking (useful for jigsaw puzzles) and non-order checking (no targets whatsoever, just blackboard with images in a random order). This enables us to allow for clues that reveal themselves as users drag around the pictures and allows a lot more freedom for admins. We can push this to another ticket. 

> thoughts on this as a final product / design flow?

Definitely not final. Design will probably need to change things, but that is a hifi issue. For now, all I care about is functionality and Vivian managed to do that. In terms of product considerations, I put it above.

@vivianliu0 Awesome work! Really impressive.

samples/ordered_puzzle/ordered_puzzle.html
vivianliu0: Oh my bad, I misinterpreted the ticket and thought that we wanted the correct answer to be unlocked/checked when placed in the correct order, so I added the targets to verify this. If we want to just have a blackboard that allows users to move around images freely and discover the hidden clue themselves, I can add a separate component to do this.

samples/ordered_puzzle/components/ordered_puzzle.js
jaydhulia: Ah I see, that makes sense. Feel free to keep this in here. SInce there is a lot of overlap between target-box and text-box, when integrating we will just use text-box with an if statement for adding text if `text != ""`.

samples/ordered_puzzle/components/ordered_puzzle.js
jaydhulia: Ah I see, that's interesting. I never knew that! Thanks for the explanation.

samples/ordered_puzzle/ordered_puzzle.html
jaydhulia: Hmmmm, I see. So in that case, in my opinion, let's keep the order checking in here as it is already implemented. Then when we are integrating with the backend code (aka moving this from `samples/*` to `app/public/*`), we will convert this into 2 different types of puzzles:
1. A puzzle with no targets and no checking. In this case, users won't have a target to go towards, but rather arranging it in the right order will reveal a clue that will help with another puzzle.
2. A puzzle with targets and checking for whether it is `solved`. This will need to be scoped out / discussed further, and have only "allowed" layouts, so we can have a pre-selected target box layout.

Does that make sense?

aaronabraham311: We can leave these for other tickets though. What we have here can be merged in, imo

jaydhulia: Yeah, that’s for a different ticket/PR for sure. I just wanted to make sure we were on the same page for future tickets of this work.

jaydhulia (approved):
Awesome looks great to me! Nice work @vivianliu0!

jaydhulia (approved):

PR #72, opened 2021-04-16
State management

jaydhulia:
- [x] Store state for all initial interactive objects
- [x] Initialize solvable objects (right now everything other than text-pane) to be unsolved
- [x] Emit an event -> change state if an object is solved
- [x] Show different behaviour if puzzle is solved or unsolved


app/public/static/js/components/animation-on-click.js
```diff
@@ -8,7 +8,18 @@ AFRAME.registerComponent("animation-on-click-run", {
   dependencies: ["animation__onclick"],
 
   init: function () {
-    var extraInfo = this.data.blackboardData;
+    const el = this.el;
+    this.data.blackboardData.jsonData.id = el.id;
+
+    const extraInfo = this.data.blackboardData;
+
+    // initialize each interactive object to not solved state, except text-panes (which are just view objects)
```
ahmedhamodi, line 16: Thanks for the comment, I was confused initially lol

app/public/templates/state.html
```diff
@@ -0,0 +1,16 @@
...
+      solvedObjects: {},
+    },
+
+    handlers: {
+      solvedObject: function (state, action) {
+        state.solvedObjects[action.id] = true;
+      },
+      initializeObject: function (state, action) {
+        state.solvedObjects[action.id] = action.solved;
+      },
```
ahmedhamodi, line 13: Is there a reason this isn't just false? Can an object be initialized in a solved state?

Also, can we not also do action.solved for solvedObject, because I'd think the solved attribute should be true.

If the above is true, then these are identical functions and perhaps can be merged into a single function.

ahmedhamodi (commented):
Code changes look good to me. Will test locally in a bit.

ahmedhamodi:
Do we want to save state on half-typed or error states?

Loom video showing what I tried to do: https://www.loom.com/share/3f6b446ef04345d18014e0cd770107e1

app/public/templates/state.html
jaydhulia: For some objects, `initializeObject` is true (for example text-panes). Other objects may also be in a solved state later on.

The reason for the different functions is because I want the code to be easily readable. Like `initializeObject` -> initializes the object. `solveObject` -> solves the object. For now, it would seem we can merge them, but keep in mind later on we will have a different flow for an object being solved vs initialized! (Aka the success flow ticket)

jaydhulia:
@ahmedhamodi , No I think for now this is okay, and we don't need to store state on half-typed answers. This can be changed if designers think we should, but for now I think this should be good. 
The reasoning: this is going to be a lot of technical work to implement that, because it cannot be generalized. Different types of puzzles (keypad vs jigsaw vs dragging) will have different type of objects that they would need to save to "restore" the state.

ahmedhamodi:
> The reasoning: this is going to be a lot of technical work to implement that, because it cannot be generalized. Different types of puzzles (keypad vs jigsaw vs dragging) will have different type of objects that they would need to save to "restore" the state.

Good point. Okay, I'm good to keep this as is.

ahmedhamodi (approved):
lgtm - tested locally too!

PR #74, opened 2021-04-16
Rotation component integration

drbalar:
Some things that still need to be figured out...

- Do we need a `scale` variable too or do we want to just use the same value used in the non-blackboard environment?
- There were two animation events. `animation-onclick` and `animation-on-click-run` which also got copied over to the cloned element because it is part of the div. So it also gets triggered when releasing the drag. So we might need a naming convention for animation events to remove it from the clone or if anyone has any other ideas


app/public/static/js/components/README.md
```diff
@@ -0,0 +1,86 @@
+# Components
```
ahmedhamodi, line 1: Ahhh thanks so much for making this readme. Love good documentation!

scripts/mockdata/objects.json
```diff
@@ -14,8 +14,16 @@
...
     "texts": ["", "hello", "from the other side"],
-    "is_interactable": false,
-    "animations_json": {}
+    "is_interactable": true,
+    "animations_json": {
+      "blackboardData": {
+        "componentType": "rotation-controls",
+        "jsonData": {
+          "position": [0, 0, 5],
+          "rotation": [0, 0, 0]
```
ahmedhamodi, line 23: Are these needed / are there times where we'd have different values here? Are we exposing this to admins?

ahmedhamodi (commented):
Left a small comment. Code changes look good to me, but will leave to Jay for final review. Will also want to test locally.

scripts/mockdata/objects.json
drbalar: not sure. Rotation we might not need but position we may or may not have to keep. If certain objects are too large then when rotating them.. they will go into the blackboard or cover the camera. So positioning maybe be needed in that case

app/public/static/js/components/rotation-controls.js
```diff
@@ -0,0 +1,100 @@
...
+
+    let childEl = document.getElementById(childId);
+    let clonedEl = document.createElement("a-gltf-model");
+    clonedEl.setAttribute("src", childEl.getAttribute("src"));
+    clonedEl.setAttribute("scale", childEl.getAttribute("scale"));
+    clonedEl.setAttribute("rotation", childEl.getAttribute("rotation"));
+
+    // TODO: Is scaling needed?
+    if (data.hasOwnProperty("position")) {
+      el.setAttribute("position", {
```
jaydhulia, line 29: Why are we setting the position for both `el` and `clonedEl`? What are the differences between the two?

jaydhulia (dismissed):

app/public/static/js/components/rotation-controls.js
drbalar: `el` is the outer entity with the rotation-controls. So that is the one rotating. `clonedEl` is the actual object nested inside `el`. `el` we set position because we need to.

And `clonedEl` position needs to be 0 0 0 otherwise it will start orbiting instead of rotating. Does that make sense?

jaydhulia: Ahhhh I see, makes sense. Is there a technical reason for having the clonedEl be separate from the outer `el`? Can we not just add the src and gltf stuff to `el` instead of `clonedEl`? Not a biggie if we can't... just wondering

drbalar: Oh truee, good point. I didnt do it originally because wasn't sure if gltf is the only stuff in there. I'll merge them into one.

jaydhulia (approved):
Nice, this looks great to me!

PR #75, opened 2021-04-17
Add scene id to a scene

drbalar:


app/routes/ui/scenario.py
```diff
@@ -62,6 +62,7 @@ async def get(self, scenario_friendly_name, scene_number=0):
             await self.render(
                 "scene.html",
                 is_last_scene=is_last_scene,
+                scene_id=scenario_obj.scene_ids[scene_number_int],
```
ahmedhamodi, line 65: nit: we are using `scenario_obj.scene_ids[scene_number_int]` in 3 places now (lines 41, 47, and 65), might be best to bring out to a variable and use that instead.

ahmedhamodi (dismissed):
Besides the small comment, lgtm

app/public/templates/scene.html
```diff
@@ -22,7 +22,8 @@
 
   <body>
     <div id="scene-root">
-      <a-scene cursor="rayOrigin:mouse" raycaster="objects: .link">
+      <!-- Need to add  inspector="url: http://localhost:3333/dist/aframe-inspector.js" to the a-scene element if using local inspector -->
+      <a-scene id="{{ scene_id }}-scene" cursor="rayOrigin:mouse" raycaster="objects: .link">
```
jaydhulia, line 26: We don't need the extra variable here, you should be able to use scene directly aka:
```
{{ scene_dict['id'] }}-scene
```

jaydhulia (changes requested):

jaydhulia (approved):

ahmedhamodi (approved):

PR #78, opened 2021-04-17
Add admin scene handler

drbalar:


app/__main__.py
```diff
@@ -60,6 +61,7 @@ def get_routes():
             UIStaticHandler,
             dict(path=f"{os.path.dirname(__file__)}/public/static/"),
         ),
+        (r"/admin/scene/([0-9]{0,16})", UIAdminSceneHandler),
```
ahmedhamodi, line 64: Does `/admin/([a-zA-Z_-]{1,50})/([0-9]{0,16})` make more sense here?

ahmedhamodi (commented):
Only one real comment, will leave to Jay for approval

app/__main__.py
drbalar: Hmmm, since we don't use the scenario friendly name anywhere in this handler i thought it would make sense to leave it out.

Your way also makes sense in a way since that matches the user endpoint.

Should I change to what you have?

ahmedhamodi: Wait for Jay's feedback first! It was just an idea

app/routes/ui/scenario.py
```diff
@@ -62,7 +62,7 @@ async def get(self, scenario_friendly_name, scene_number=0):
             await self.render(
                 "scene.html",
                 is_last_scene=is_last_scene,
-                scenario_dict=scenario_obj.as_dict(),
+                scenario_name=scenario_obj.name,
```
jaydhulia, line 65: Let's add the `is_admin=False` here

app/public/templates/scene.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>{{ scenario_dict["name"] }} - {{ scene_dict["name"] }}</title>
+    <title>{{ scenario_name }} - {{ scene_dict["name"] }}</title>
```
jaydhulia, line 4: I can't add a comment on line 25-26, but this is for that:
Since we are adding the `is_admin` parameter, can we use that for that line similar to camera (https://github.com/uwblueprint/cdc-backend/blob/master/app/public/templates/scene.html#L78) for adding inspector:
```
<a-scene id="{{ scene_dict['id'] }}-scene" cursor="rayOrigin:mouse" raycaster="objects: .link"
{% if is_admin %}
            inspector="url: {{ inspector_url }}"
 {% end %}
>
```
And then in the python admin scenario handler:
```
inspector_url=config.get("inspector_url")
```
And add this one to dev-config:
inspector_url: http://localhost:3333/dist/aframe-inspector.js

That way, when switching to prod, we just have to update the config and the rest will work out.

app/routes/ui/admin_scene.py
```diff
@@ -0,0 +1,48 @@
...
+
+    async def get(self, scene_id=1):
+        try:
+            scene_id_int = int(scene_id)
+
+            self.db_session = get_session()
+            scene_dict = await get_scene_from_postgres(
+                scene_id_int,
+                self.db_session,
+                update_cache=True,
```
jaydhulia, line 22: Let's set update_cache=False here (the reason being, we don't touch the cache on admin routes as they may make a lot of updates in a short period of time, so no need to access or update the cache to reduce latency).

app/public/templates/scene.html
drbalar: 🧠

jaydhulia (approved):
Awesome, thanks for making the changes! LGTM!

jaydhulia (approved):

PR #79, opened 2021-04-18
Allow arbitrary length passwords from admin

jaydhulia:


drbalar (approved):

app/public/static/js/components/keypad.js
```diff
@@ -18,8 +18,10 @@ AFRAME.registerComponent("keypad", {
 
     el.setAttribute(
       "super-keyboard",
-      "imagePath:/static/img/; multipleInputs:true; align: center; maxLength: 4; model:" +
-        data.model
+      "imagePath:/static/img/; multipleInputs:true; align: center; model:" +
+        data.model +
+        "; maxLength: " +
+        password.length
```
drbalar, line 24: It is irrelevant to the current task but one optional change you can try is with json like structure just for readability .. so like:
```
el.setAttribute(
      "super-keyboard",
      {
        "imagePath": "/static/img/",
        "multipleInputs":true,
        "align": "center",
        "model": "data.model",
        "maxLength": password.length
      }
    );```

drbalar (approved):

PR #80, opened 2021-04-20
GET all endpoints

jaydhulia:
- [x] GET all assets
- [x] GET all scenarios
- [x] GET all scenes

### DIFF
--- a/app/library/postgres.py
+++ b/app/library/postgres.py
@@ -7,6 +7,7 @@ from models.db_client import (
     delete_scenario,
     delete_scene,
     get_asset,
+    get_assets,
     get_object,
     get_scenario,
     get_scenario_by_friendly_name,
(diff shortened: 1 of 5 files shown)


app/library/postgres.py
```diff
@@ -86,6 +95,12 @@ async def get_asset_from_postgres(asset_id: str, session, update_cache=False):
     return response
 
 
+async def get_assets_from_postgres(session):
+    assets = get_assets(session)
+    resp_dict = {"assets": [asset.as_dict() for asset in assets]}
+    return resp_dict
```
aaronabraham311, line 101: Specifically, which assets would this get? Is it both scene and object assets?

jaydhulia: Well neither. Scene doesn’t have assets, and objects are just linked to scenes.
These will return the list of actual assets that we have, like the objects in S3. Like in the sample escape room, this would be the models chair, living room, classroom etc.

aaronabraham311: So when we want to see assets on the frontend, how do we show the admin that a certain asset is a scene asset versus an object asset? Furthermore, in the backend UML diagram, there seems to be a relation between scenes and assets, so don't scenes technically have assets?

aaronabraham311: Clarified on Slack. All good

jaydhulia: Resolved offline on Slack

ahmedhamodi (approved):
lgtm

drbalar (approved):
tried them out, lgtmt!

PR #81, opened 2021-04-24
Success flow for puzzles

jaydhulia:
- [x] Close popup after X seconds, on a puzzle being solved
- [x] Navigate to next (or current) scene on last object being solved

### DIFF
--- a/app/routes/ui/admin_scene.py
+++ b/app/routes/ui/admin_scene.py
@@ -35,7 +35,6 @@ class UIAdminSceneHandler(BaseUIHandler):
                 "scene.html",
                 is_last_scene="false",
                 is_admin=True,
-                scene_id=scene_id_int,
                 scenario_name="Edit Scene",
                 scenario_friendly_name="friendly_name",
                 scene_dict=scene_dict,
--- a/app/routes/ui/admin_scene.py
+++ b/app/routes/ui/admin_scene.py
@@ -33,15 +33,17 @@ class UIAdminSceneHandler(BaseUIHandler):
 
             await self.render(
                 "scene.html",
-                is_last_scene=False,
+                is_last_scene="false",
                 is_admin=True,
                 scene_id=scene_id_int,
                 scenario_name="Edit Scene",
+                scenario_friendly_name="friendly_name",
                 scene_dict=scene_dict,
                 asset_prefix_url=config.get("asset.prefix_url"),
                 navmesh_src=navmesh_src,
                 inspector_url=config.get("inspector_url"),
                 json=json,
+                cur_scene_idx=scene_id_int,
             )
         except ValueError as e:
             self.write_error(status_code=404, message=str(e))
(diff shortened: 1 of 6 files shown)


app/routes/ui/admin_scene.py
```diff
@@ -33,15 +33,17 @@ async def get(self, scene_id):
 
             await self.render(
                 "scene.html",
-                is_last_scene=False,
+                is_last_scene="false",
```
ahmedhamodi, line 36: Any reason for string over the actual boolean value?

app/routes/ui/admin_scene.py
```diff
@@ -33,15 +33,17 @@ async def get(self, scene_id):
...
                 is_admin=True,
                 scene_id=scene_id_int,
                 scenario_name="Edit Scene",
+                scenario_friendly_name="friendly_name",
                 scene_dict=scene_dict,
                 asset_prefix_url=config.get("asset.prefix_url"),
                 navmesh_src=navmesh_src,
                 inspector_url=config.get("inspector_url"),
                 json=json,
+                cur_scene_idx=scene_id_int,
```
ahmedhamodi, line 46: Why is this field needed?

ahmedhamodi (commented):
Overall lgtm, couple of comments.

It seems to me that the new `cur_scene_idx` field isn't needed since it never gets updated? So can just use `scene_id` field unless I'm missing something.

app/routes/ui/admin_scene.py
jaydhulia: `scene_id` is not the same as `cur_scene_idx`. `scene_id` is the id of the scene in postgres, while `cur_scene_idx` is the index of the current scene.
Example:
`scenario: [5, 77, 32, 23]`, then if someone opens the scenario: `scene_id` would be 5, and `cur_scene_idx` would be 0.

app/routes/ui/admin_scene.py
jaydhulia: Javascript vs Python boolean. `False` is an undefined variable in HTML/JS, whereas "false" works

app/routes/ui/admin_scene.py
ahmedhamodi: On lines 38 and 46, the two fields `scene_id` and `cur_scene_idx` are set to the same thing, so is this a mistake then?

jaydhulia: Yeah, that wasn't me and was leftover from another PR (https://github.com/uwblueprint/cdc-backend/pull/78) by accident. I can remove it, but it's not a biggie.

jaydhulia: Removed the unused variable `scene_id`

ahmedhamodi: Ah thank you. Makes sense now!

ahmedhamodi (approved):

PR #82, opened 2021-04-24
Removing CORS errors for GET requests

aaronabraham311:


app/routes/base.py
```diff
@@ -34,6 +36,10 @@ class BaseAdminAPIHandler(BaseAPIHandler):
     Base handler for all api/admin/* routes
     """
 
+    async def options(self):
+        self.set_status(204)
+        await self.finish()
```
ahmedhamodi, line 41: @jaydhulia - Do we want to do any validation here and return a prettier error as opposed to CORS standard error message?

jaydhulia: `options` is typically not supposed to return any errors. It will just return an OK and the actual route will take care of error handling.

app/routes/base.py
```diff
@@ -15,6 +15,8 @@ def prepare(self):
 
     def set_default_headers(self) -> None:
         self.set_header("Content-Type", "application/json")
+        self.set_header("Access-Control-Allow-Origin", "*")
```
jaydhulia, line 18: Added to look into this more for pre-deployment checklist

jaydhulia (approved):

jaydhulia:
Moved CORS stuff to admin only routes, since only they need it @aaronabraham311

PR #84, opened 2021-04-25
Adding content type to allowed headers

aaronabraham311:
In reference to [this conversation](https://github.com/uwblueprint/cdc-frontend/pull/36#discussion_r619837769)

### DIFF
--- a/app/routes/base.py
+++ b/app/routes/base.py
@@ -15,9 +15,6 @@ class BaseAPIHandler(tornado.web.RequestHandler):
 
     def set_default_headers(self) -> None:
         self.set_header("Content-Type", "application/json")
-        self.set_header(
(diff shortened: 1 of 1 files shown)


app/routes/base.py
```diff
@@ -15,6 +15,9 @@ def prepare(self):
 
     def set_default_headers(self) -> None:
         self.set_header("Content-Type", "application/json")
+        self.set_header(
```
jaydhulia, line 18: Instead of adding it here (this is the general base method), let’s modify the allowed header on admin routes only (on line 47)

jaydhulia (changes requested):

jaydhulia (approved):

PR #85, opened 2021-04-26
Allowing arbitrary options arguments and specifying methods

aaronabraham311:
- PUT requests from Axios issues an OPTIONS request before the PUT requests. This causes two problems:
    - Our current base implementation of OPTIONS only allows one argument, but we did not account for requests that have multiple arguments (as is the case for PUT)
    - Axios will only allow requests from the Access-Control-Allow-Methods that is returned from OPTIONS, which currently bars PUT requests

- The above issues are resolved in this PR (discussed with @jaydhulia)

### DIFF
--- a/app/routes/base.py
+++ b/app/routes/base.py
@@ -34,7 +34,7 @@ class BaseAdminAPIHandler(BaseAPIHandler):
     Base handler for all api/admin/* routes
     """
 
-    async def options(self):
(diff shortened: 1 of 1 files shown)


jaydhulia (approved):
Assuming that you tested this already with the PUT request on FE :) Thanks for the PR @aaronabraham311

PR #86, opened 2021-04-27
Jigsaw Puzzle Component

vivianliu0:
Loom Demo Video: 
https://www.loom.com/share/effd2dac6f2d43eea1ae5000f6de3045


ahmedhamodi (commented):
Purely from a design perspective, I think it would look better if there was a gap between the grid positions, thoughts @aaronabraham311?

vivianliu0:
Updated the spacing of the jigsaw puzzles.
Capture

samples/jigsaw_puzzle/jigsaw-prototype.py
```diff
@@ -1,6 +1,6 @@
 import cv2
```
jaydhulia, line 1: Note: to run this, need to install extra lib `pip install opencv-python`

jaydhulia (approved):
Nice work @vivianliu0, looks great to me!

samples/jigsaw_puzzle/jigsaw-prototype.py
vivianliu0: Added a comment in the python file

jaydhulia (approved):

PR #87, opened 2021-04-27
Added a sample visual pane demo in visual_pane.html

drbalar:
https://www.loom.com/share/382705afbbb4494c8bc9074889a828cf


ahmedhamodi (commented):
Mentioned this on slack, but also will mention here:

In terms of exposing the text position, I like the “above” vs “below” but I don't think the other fields should be exposed to an admin. Thoughts @aaronabraham311?

Also, can we make the image larger? What's the limiting factor on the image — is it the actual image itself or are should we be scaling it before displaying it?

drbalar:
Update after the video^.. `width` and `height` are gone. Instead there is a `scaleBy` field to adjust the size of the image which is scaled while maintaining its aspect ratio

aaronabraham311:
> In terms of exposing the text position, I like the “above” vs “below” but I don't think the other fields should be exposed to an admin. Thoughts

I don't think the above or below option should be surface to admins. I don't see much of a use case here for admins. If it turns out during user testing that this is actually needed, then we can enable it. Would prefer to keep the admin interface as simple as possible. However, it's awesome that you implemented this Dhruvin! We have this in case we need it. 

Do we define some sort of default height and width that we scale an image to? What happens if an admin has a wonky image?

drbalar:
replied on slack.. but
We use the ratio of the image so they have to insert a normal image otherwise it will scale the wonky image. For the default width/height.. we dont have anything but we can define something like height = 3. And width will auto-adjust or vice-versa

samples/visual_pane/components/visual_pane.js
```diff
@@ -0,0 +1,71 @@
...
+    this.imageEl.setAttribute("class", "link");
+
+    this.el.appendChild(this.imageEl);
+
+    this.textEl = document.createElement("a-text");
+    this.textEl.setAttribute("id", "text");
+    this.textEl.setAttribute("value", data.text);
+    this.textEl.setAttribute("negate", "true");
+    this.textEl.setAttribute("scale", "2 2 1");
+    let textVerticalOffset = 1;
```
jaydhulia, line 42: nit: ternary operators exist in JS 😄 
```
const textVerticalOffset = data.hasOwnProperty("textVerticalOffset") ? data.textVerticalOffset : 1;
```

samples/visual_pane/components/visual_pane.js
```diff
@@ -0,0 +1,71 @@
...
+
+  init: function () {
+    const data = this.data.jsonData;
+    const el = this.el;
+
+    this.imageEl = document.createElement("a-image");
+    this.imageEl.setAttribute("id", "puzzle-piece-image");
+    this.imageEl.setAttribute("src", data.imageSrc);
+
+    const rawImageEl = document.querySelector(data.imageSrc);
```
jaydhulia, line 19: Ahhh, this may be problematic when integrating with backend. In the actual backend code, this is going to be a direct img URL (to our S3), rather than a local reference (#) to a-asset. Can you update the HTML's imgSrc to directly be `./pic2.jpg` and then still get this to work?

samples/visual_pane/visual_pane.html
```diff
@@ -0,0 +1,42 @@
...
+    <script src="components/visual_pane.js"></script>
+  </head>
+  <body>
+    <a-scene
+      background="color: #AAD5D7"
+      cursor="rayOrigin: mouse"
+      raycaster="objects: .link"
+    >
+      <a-assets>
+        <img id="visual1" src="./pic2.jpg" crossorigin="anonymous" />
```
jaydhulia, line 15: `pic2.jpg` isn't included in this PR, but `pic1.png` is, does this need to be updated? 😄

jaydhulia:
@drbalar The text doesn't show up for me on running the preview:
image
The only thing I changed was the src to './pic1.png`

samples/visual_pane/visual_pane.html
drbalar: oh yes, ill change back to `pic1.png`

samples/visual_pane/components/visual_pane.js
drbalar: Ah I see what you mean. To get the image dimensions I had to load it into an `` tag first and then read from it. I'll see if there is a way to read directly without creating img tag in that case.

drbalar: I am not able to find a very good alternate way .. one way I saw was using javascript to create an `Image` object, and then doing the work in its `onLoad` function but there were comments saying it might not always work depending on browser and version..

What do you think about preloading this `img` tag under Aframe-assets as part of the server-side rendering?

jaydhulia: I would prefer to avoid doing special things for a component, if possible, so we can have a generic way for each component. Otherwise, we would have to loop through all objects every time. I can think of a couple options that we can explore:
1. Is the only thing you need, the image's width and height? Are these based on the original image upload? If they are, then we can just add them to jsonData and add the width/height of the image as part of the logic that saves the image originally. So, on the admin interface, admin would upload the image and when we are saving it as part of the scene, we automatically append the width and height. Would this work?
2. I believe we should be able to add an `` dynamically using JS? Like:
```
var img = document.createElement("img");
img.src = "http://www.google.com/intl/en_com/images/logo_plain.png";
```
and then append it to the HTML page if needed:
```
var src = document.getElementById("header");
src.appendChild(img);
```
Since the `` tag will exist on the page you should be able to get the width and height I believe.

Let me know if any of the above need more clarifications, or if both solutions above won't work, in which case I will be good with adding conditionals on serverside rendering 😛

drbalar: Yes, the width and height is all that is needed. This is just to figure out the ratio of the image. I like both suggestions.

I tried out `option 2`. It seems to work fine at the moment.. I guess we can switch to 1 if we encounter any issues

samples/visual_pane/components/visual_pane.js
```diff
@@ -0,0 +1,74 @@
...
+    imageEl.setAttribute("src", data.imageSrc);
+
+    let rawImageEl = document.createElement("img");
+    rawImageEl.src = data.imageSrc;
+    rawImageEl.onload = function () {
+      const ratio = rawImageEl.width / rawImageEl.height;
+      const scaleBy = data.hasOwnProperty("scaleBy") ? data.scaleBy : 5;
+      imageEl.setAttribute("width", scaleBy * ratio);
+      imageEl.setAttribute("height", scaleBy);
+      // Randomizes the position of the puzzle piece on the blackboard
```
jaydhulia, line 24: This is a leftover comment, right?

samples/visual_pane/components/visual_pane.js
```diff
@@ -0,0 +1,74 @@
...
+
+  init: function () {
+    const data = this.data.jsonData;
+    const el = this.el;
+    let imageEl = document.createElement("a-image");
+    imageEl.setAttribute("src", data.imageSrc);
+
+    let rawImageEl = document.createElement("img");
+    rawImageEl.src = data.imageSrc;
+    rawImageEl.onload = function () {
```
jaydhulia, line 19: I think the `.onload` needs to be before adding the `.src` so I would switch it around. Because otherwise the image may load before the onload listener is defined

drbalar: good catch

jaydhulia (approved):
Awesome, thanks for the changes, looks good to me! 🎉

PR #89, opened 2021-04-30
Image pane component integration into backend

drbalar:
If you are running this..
- I added a another chair to scene 3 (classroom). It is at the front of the classroom.
- You will need to update the database as mockdata has been changed.

Also updated component docs with visual pane mockdata info


docs/components-README.md
```diff
@@ -82,3 +82,33 @@ Simple documentation for components being used and a sample of their `animations
   ...
 }
 ```
+
+## Visual Pane
+
+- `animations_json` - object
+  - `blackboardData`- object
+    - `componentType` - Needs to be the string "visual-pane"
```
ahmedhamodi, line 90: Can we list the possible values for componentType somewhere in this readme?

ahmedhamodi:
Looks good to me - thanks for readme update.

Def want Jay to review still before approving.

docs/components-README.md
jaydhulia: This is a good idea; once we have all the components we need, I will add an enum validation to the backend for ensuring it is only one of those when objects are modified.

app/public/static/js/components/visual-pane.js
```diff
@@ -0,0 +1,76 @@
...
+      if (data.hasOwnProperty("position")) {
+        el.setAttribute("position", {
+          x: data.position[0],
+          y: data.position[1],
+          z: data.position[2],
+        });
+      } else {
+        el.setAttribute("position", { x: 0, y: 0, z: 0 });
+      }
+      imageEl.setAttribute("class", "link");
```
jaydhulia, line 33: The imageEl shouldn't be clickable :)

jaydhulia (approved):
One small comment, but looks great to me! Nice work @drbalar

jaydhulia (approved):

PR #90, opened 2021-04-30
Integrate hints

jaydhulia:
- Hints are being pulled from Postgres for the scene and displayed properly
- New Hints and Hints button
- Button is always on the right bottom corner of the screen

On initial loading of page:
image

Once you have clicked on the hint at least once:
image


app/public/templates/scene.html
```diff
@@ -81,11 +82,22 @@
...
           position="0 {{ scene_dict['camera_properties']['position'][1] }} 0"
-        ></a-entity>
+        >
+        <!--   Hints    -->
+        <a-entity
+          id="hint-button"
+          class="link"
+          hint-button='jsonData: {"newHintImg": "/static/img/lightOn.png", "hintImg": "/static/img/lightOff.png", "width": 0.08, "height": 0.08}'
+          animation__onclick="property: position; dur: 0; startEvents: click"
+          animation-on-click-run='blackboardData: {"jsonData": {"text": {% raw json.dumps(scene_dict["hints"]) %}, "currPosition": 0}, "componentType": "text-pane"}'
```
ahmedhamodi, line 97: Should we pass back a default value for "text" when there are no hints (ie. empty array)? Something like "no hints at the moment" would make sense to me.

app/public/templates/scene.html
```diff
@@ -81,11 +82,22 @@
         <a-entity
           id="primaryCamera"
           camera
+          fov="80"
```
ahmedhamodi, line 85: Curiosity: any particular reason to change the field of view? Also is 90 default?

ahmedhamodi (commented):
Small comments, otherwise looks good to me!

app/public/templates/scene.html
jaydhulia: Yes, this will be done on the backend once designers have the right lingo ironed out. The default will be updated here: https://github.com/uwblueprint/cdc-backend/blob/master/app/models/scene.py#L18. For now, that's why I have put `generic hint for all scenes` as the first element in the mockdata. There will always be at least one hint.

app/public/templates/scene.html
jaydhulia: I am not changing the field of view, I am just making it visible :) 80 is the default for aframe: https://aframe.io/docs/1.1.0/components/camera.html#properties

ahmedhamodi (approved):
lgtm!

app/public/static/js/components/hint-button.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+    this.el.appendChild(this.hintIcon);
+
+    // Update the location of the hint button on the screen depending on the screen size
+    window.addEventListener("resize", function () {
+      const z = -0.5;
+      const x = visibleWidthAtZDepth(z) / 2 - data.width;
+      const y = -visibleHeightAtZDepth(z) / 2 + data.height;
+
+      hintIcon.object3D.position.set(x, y, z);
```
drbalar, line 50: How does this differ from the
```
hintIcon.setAttribute("positiion", {...});
```
way? This modifies a different `position` field?

drbalar (approved):
Looks good :D , just a question

app/public/static/js/components/hint-button.js
jaydhulia: Yeah, it’s modifying the THREE.J’s’s position field rather than the aframe object position field. I think this is needed because we are using math to figure out the bottom corner of the screen based on the THREE.js values

drbalar: Ah I see

PR #91, opened 2021-05-01
Add Name Field to Object

vivianliu0:
Add object name field to object schema (and update mock data, validation, etc)


ahmedhamodi (commented):
Looks good to me.

~~@jaydhulia - do we need to update any schema validation to include this new field?~~
Edit: I'm blind, it is there. Is the 1-50 character limit good?

jaydhulia:
@ahmedhamodi Yeah I think 1-50 characters should be more than enough for object names. We can increase this later, if needed. 
@vivianliu0 Looks good to me! I just updated the regex to also allow numbers, because people may want like `Teacup 1` `Teacup 2` etc.

jaydhulia (approved):

jaydhulia:
I merged it because I need to fix my other PR based on these changes 😄

PR #92, opened 2021-05-01
Integrate target and non-target puzzles into backend

jaydhulia:
Created the objects, saved state (for puzzles with targets) etc.
Two chairs on the front, the left chair is puzzle with targets, right chair is puzzle without targets.
image

Puzzle with targets initial state:
image
Puzzle with targets solved state:
image

Puzzle with no targets:
image


app/public/static/js/components/ordered-puzzle.js
```diff
@@ -0,0 +1,196 @@
...
+        depth: "1",
+        color: "green",
+        x: "0",
+        y: "-5.75",
+        z: "0",
+        scaleX: "4",
+        scaleY: "4",
+        scaleZ: "2",
+        text: "Solved",
+      };
```
ahmedhamodi, line 38: Correct me if I'm wrong, but these attributes are going to be the same for all solved puzzle texts right? (same position, color, scale, etc). Should we try to extract this to a generic place?

app/public/static/js/components/ordered-puzzle.js
```diff
@@ -0,0 +1,196 @@
...
+          );
+          puzzlePiece.setAttribute("onTarget", false);
+
+          puzzlePiece.addEventListener("dragend", function (event) {
+            event.target.setAttribute("onTarget", isOnTarget(event.target));
+            if (isPuzzleComplete(puzzlePieceCache)) {
+              // emit event to update state
+              el.sceneEl.emit("solvedObject", { id: data.id });
+              // emit event to close popup (after 1 second for now) - TODO: disabled for now
+              // need designer input on whether we should autoclose this
```
ahmedhamodi, line 137: Discussed offline, but my thought is:
- should not autoclose as these puzzles might need some time to derive the meaning / clue from
- even if it's the last puzzle, we should not move them into the next room until the popup is closed

app/public/static/js/components/ordered-puzzle.js
```diff
@@ -0,0 +1,196 @@
...
+    this.puzzleIsSolved = this.data.isSolved[this.id];
+
+    if (typeof this.puzzleIsSolved === "undefined") {
+      // not loaded yet, do nothing
+    } else if (this.puzzleIsSolved === true && this.useTargets) {
+      // Already solved
+      this.solvedPuzzleEntity.setAttribute("visible", "true");
+    } else {
+      // not solved yet, do nothing for now
+    }
```
ahmedhamodi, line 171: Nit: can just do the `else if` clause as the `if` and then both the not loaded and not solved can be caught with a single `else`

ahmedhamodi (commented):
Some thoughts -- @drbalar can you review as well?

app/public/static/js/components/ordered-puzzle.js
jaydhulia: Unfortunately, this won’t be the same for all of them, the scaling, xyz and height/width are different for this vs keypad for example.

app/public/static/js/components/ordered-puzzle.js
jaydhulia: Yeah I noticed that, I left it this way for the comments in the different cases; right now we don’t do anything specific but in the future we will :) so leaving it here for future references.
As an example, in the last else, we can randomize position instead of init/ that way, we only randomize if it hasn’t been solved. Of course this is for the future. I can change it for this PR you want

app/public/static/js/components/ordered-puzzle.js
ahmedhamodi: Ah rip, okay

app/public/static/js/components/ordered-puzzle.js
ahmedhamodi: Nah this is a good point -- good to leave as is.

app/public/static/js/components/ordered-puzzle.js
```diff
@@ -0,0 +1,196 @@
...
+});
+
+function isOnTarget(puzzlePiece) {
+  const delta = puzzlePiece.object3D.children[0].position;
+  const originalPos = puzzlePiece.getAttribute("position");
+  const targetPos = {
+    x: puzzlePiece.getAttribute("xTarget"),
+    y: puzzlePiece.getAttribute("yTarget"),
+  };
+  const accuracyReq = 1;
```
drbalar, line 182: Can we change this to something like?
```
const accuracyReq = scaleBy/3;
```

Otherwise at lower scaling I am able get solved state in such situations:
image

drbalar (approved):
Looks good :D

PR #94, opened 2021-05-06
Jigsaw Component Integration

vivianliu0:
Added a desk at the front of the classroom that when clicked, allows the user to interact with a jigsaw puzzle. 

https://www.loom.com/share/755f7621ebd5421786bc91a90b0660fb


app/public/static/js/components/jigsaw-puzzle.js
```diff
@@ -0,0 +1,86 @@
...
+      "ordered-puzzle",
+      "jsonData",
+      JSON.stringify(ordered_puzzle_json)
+    );
+    this.ordered_puzzle.setAttribute("isSolved", this.data.isSolved);
+    this.el.appendChild(this.ordered_puzzle);
+  },
+
+  update: function () {
+    this.puzzleIsSolved = this.data.isSolved[this.id];
```
jaydhulia, line 64: Ah this doesn't persist if you close and re-open the puzzle, you will see that the `Solved` part is gone. Instead what we should do is just use the same binding that we use for ordered-puzzle normally.

So you can remove this update method entirely (see comment on line 59)

app/public/static/js/components/jigsaw-puzzle.js
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    this.ordered_puzzle = document.createElement("a-entity");
+    this.ordered_puzzle.setAttribute("id", "ordered-puzzle");
+    this.ordered_puzzle.setAttribute(
+      "ordered-puzzle",
+      "jsonData",
+      JSON.stringify(ordered_puzzle_json)
+    );
+    this.ordered_puzzle.setAttribute("isSolved", this.data.isSolved);
```
jaydhulia, line 59: Instead of adding a static map here, we should add a binding that will automatically call ordered-puzzle's update method:
```
    this.ordered_puzzle.setAttribute(
    "bind__ordered-puzzle",
    "isSolved: solvedObjects"
    );
```
that way, any time `isSolved` is updated, this will automatically call ordered-puzzle's update method. By calling its update method, jigsaw doesn't have to worry about toggling visibility of the "Solved" thingy.

scripts/mockdata/objects.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "name": "Name1",
+    "name": "Classroom rotation inspector",
```
jaydhulia, line 3: Thanks for updating the names 😄  This was very much needed lol

jaydhulia (commented):
Looks great overall! Just need to address the persistence issue.

aaronabraham311:
Looks awesome from a design/product perspective, Vivian. Great work as always.

app/public/static/js/components/ordered-puzzle.js
```diff
@@ -145,27 +145,24 @@ AFRAME.registerComponent("ordered-puzzle", {
...
-          puzzlePiece.setAttribute(
-            "onTarget",
-            isOnTarget(puzzlePiece, scaleBy)
-          );
-        });
-      });
-    }
   },
   update: function () {
+    console.log("updating");
```
jaydhulia, line 163: Can we get rid of the console.log statements? Looks good otherwise, will test locally later!

app/public/static/js/components/jigsaw-puzzle.js
vivianliu0: Ooh I see, thanks for explaining, updated the PR

jaydhulia (approved):
Looks good to me!

PR #95, opened 2021-05-15
Added `transitions` column to `scenarios` table.

drbalar:
Updated:
- mockdata
- create-tables.py
- api_schemas.py
- scenario.py

Also added a default value in dev-config to use when creating a new scenario.


configs/dev-config.yaml
```diff
@@ -21,6 +21,21 @@ asset:
...
+          [
+            {
+              "text": "INTRODUCTION: text goes here.",
+              "imageSrc": "image goes here",
+            },
+          ],
+      },
+      { "data": [{ "text": "CONCLUSION: just text, no images" }] },
+    ]
+
```
ahmedhamodi, line 38: Interesting that we have this in config, do we do this for any other fields? Because for now we can set mock data to have it, and then always assume (since we can control this) that all environments have these fields set.

ahmedhamodi (commented):
One quick comment, otherwise lgtm.

Also, let's update the branch since it's behind master at the moment

configs/dev-config.yaml
drbalar: Talked with @jaydhulia about this, when we deploy the application.. when CDC has finalized the default text for intro and conclusion, we can update the config file and not have to modify the code.

But yeah for dev purposes.. it would not make a difference since we have our mockdata

configs/dev-config.yaml
```diff
@@ -21,6 +21,21 @@ asset:
...
+    [
+      {
+        "data":
+          [
+            {
+              "text": "INTRODUCTION: text goes here.",
+              "imageSrc": "image goes here",
+            },
+          ],
+      },
```
jaydhulia, line 35: Similarly, let's add the `currPosition` to the default. Basically, if we were to use this default, we want to ensure nothing breaks.

scripts/mockdata/scenarios.json
```diff
@@ -8,7 +8,20 @@
     "is_previewable": true,
     "publish_link": null,
     "preview_link": "www.previewlink.com/student-escape-room",
-    "expected_solve_time": "10 to 20"
+    "expected_solve_time": "10 to 20",
+    "transitions": [
+      {
+        "data": [
+          { "text": "text and an image", "imageSrc": "bananaaaaaaaaaa" },
```
jaydhulia, line 15: Valid imageSrc 🙏  so that we can integrate easily. So similar to the other objects, something like `/static/img/lightOn.png`

scripts/mockdata/scenarios.json
```diff
@@ -8,7 +8,20 @@
...
+          {
+            "text": "text and an another image",
+            "imageSrc": "moreBananaaaaaaaaaa"
+          }
+        ],
+        "currentPosition": 0
```
jaydhulia, line 21: `currPosition` for consistency

jaydhulia (approved):
LGTM! Thanks @drbalar

PR #97, opened 2021-05-17
Conditional Image Rendering for Text Panes

vivianliu0:
Added support for conditional image rendering, which will be used by transitions.

Loom demo:
https://www.loom.com/share/d0545776713143d8a2ef5ebedd1157ea


app/public/templates/scene.html
```diff
@@ -99,7 +99,7 @@
           class="link"
           hint-button='jsonData: {"newHintImg": "/static/img/lightOn.png", "hintImg": "/static/img/lightOff.png", "width": 0.08, "height": 0.08}'
           animation__onclick="property: position; dur: 0; startEvents: click"
-          animation-on-click-run='blackboardData: {"jsonData": {"text": {% raw json.dumps(scene_dict["hints"]) %}, "currPosition": 0}, "componentType": "text-pane"}'
+          animation-on-click-run='blackboardData: {"jsonData": {"data": {% raw json.dumps( scene_dict["hints"] ) %}, "currPosition": 0}, "componentType": "text-pane"}'
         >
```
vivianliu0, line 103: Because of the schema changes that were made to the text-pane component, the hints pane isn't working at the moment since the jsonData is not following the correct schema. I was confused about the `{% raw json.dumps(scene_dict["hints"]) %}` part - is the `{% %}` a template engine system based on jinja?

Old Schema:
`{"text": ["hi", "hello"], "currPosition":0}`

New Schema:
`{"data": [{"text": "hi"}, {"text": "hello", "imageSrc": "path"}], "currPosition": 0}`

jaydhulia: Ah yes, `{% %}` is part of the tornado template system. Everything inside of it is _basically_ Python. 
So, what we were doing before is using the json library to dump the array in JS friendly manner (hence `raw`).
I fixed it by changing it to:
`{% raw json.dumps( [ {"text": x} for x in scene_dict["hints"] ] ) %}`.
List comprehension in Python,
so now `["generic hint for all scenes", "scene specific hint 1"]` will become `[{'text': 'generic hint for all scenes'}, {'text': 'scene specific hint 1'}]` and then we json.dumps it.
Overall this will give what the text-pane now expects!

jaydhulia (approved):
Looks good to me! Thanks @vivianliu0

jaydhulia:
@aaronabraham311 as a side note, in the future do you think hints will need a text + image? Or just text is okay? Right now, only text is supported. If you want hints to also be able to support images, please let me know and we can work on it in the future.

aaronabraham311:
@vivianliu0 Looks great!

@jaydhulia For now let's keep it with just textual hints. I don't really see an urgent need to allow hint pictures

drbalar (approved):
Works nice, looks good to me too!

PR #100, opened 2021-05-19
fix description regex

jaydhulia:


amoliksingh (approved):

app/library/api_schemas.py
```diff
@@ -20,7 +20,7 @@
     "properties": {
         "name": {"type": "string", "pattern": r"^[a-zA-Z _-]{1,50}$"},
         "friendly_name": {"type": "string", "pattern": r"^[a-zA-Z_-]{1,50}$"},
```
ahmedhamodi, line 22: Can we also add 0-9 on names and friendly names?

PR #101, opened 2021-05-19
Fix delete scene request + mock data issues on object level

ahmedhamodi:
When a scene was deleted it was not also removed from the scene_ids array on the scenario level. So when a subsequent fetch was made for the scenes, we would get an error. Also, the mock data for objects was not properly setup since we do not allow same objects in multiple rooms, so when a scene is deleted so are all the nested objects, but then that causes issues for the scene since its objects are deleted. Made them independent on mock data.

### DIFF
--- a/app/library/postgres.py
+++ b/app/library/postgres.py
@@ -213,11 +213,10 @@ async def delete_scene_from_postgres(scene_id: str, session):
     scenarios = get_scenarios(session)
     for scenario in scenarios:
         if int(scene_id) in scenario.scene_ids:
-            scenario_obj: Scenario = await get_scenario_from_postgres(
-                scenario.id, session
+            scenario.scene_ids.remove(int(scene_id))
+            await update_scenario_from_postgres(
+                scenario.id, scenario.as_dict(), session
             )
-            scenario_obj["scene_ids"].remove(int(scene_id))
-            await update_scenario_from_postgres(scenario.id, scenario_obj, session)
 
     response = {"message": "Deleted successfully"}
     return response
--- a/app/library/postgres.py
+++ b/app/library/postgres.py
@@ -209,6 +209,16 @@ async def delete_scene_from_postgres(scene_id: str, session):
     if not delete_scene(scene_id, session):
         raise ValueError("Scene ID not valid")
 
+    # Remove the scene from scenario's scene_ids array
+    scenarios = get_scenarios(session)
+    for scenario in scenarios:
+        if int(scene_id) in scenario.scene_ids:
+            scenario_obj: Scenario = await get_scenario_from_postgres(
+                scenario.id, session
+            )
+            scenario_obj["scene_ids"].remove(int(scene_id))
+            await update_scenario_from_postgres(scenario.id, scenario_obj, session)
+
     response = {"message": "Deleted successfully"}
     return response
 
--- a/scripts/mockdata/objects.json
+++ b/scripts/mockdata/objects.json
@@ -301,5 +301,143 @@
         }
       }
     }
+  },
+  {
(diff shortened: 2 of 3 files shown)


scripts/mockdata/scenes.json
```diff
@@ -26,7 +26,7 @@
   {
     "name": "Living Room",
     "description": "A standard master bedroom. King size bed, a dresser, ensuit washroom, and a large closet.",
-    "object_ids": [2],
+    "object_ids": [12],
```
jaydhulia, line 29: This should still be 2, 2 is the sanitizer in the Living room.

scripts/mockdata/objects.json
```diff
@@ -301,5 +301,143 @@
...
+        "jsonData": {
+          "password": "abcde",
+          "model": "basic",
+          "is_last_object": true
+        }
+      }
+    }
+  },
+  {
+    "name": "Classroom keypad",
```
jaydhulia, line 402: This object isn't needed

scripts/mockdata/objects.json
```diff
@@ -301,5 +301,143 @@
...
+        "jsonData": {
+          "password": "1234",
+          "model": "numpad",
+          "is_last_object": true
+        }
+      }
+    }
+  },
+  {
+    "name": "Classroom keyboard",
```
jaydhulia, line 374: Can you rename the objects to map appropriately? For example, this is the Student bedroom keyboard, not the classroom keyboard.

app/library/postgres.py
```diff
@@ -209,6 +209,16 @@ async def delete_scene_from_postgres(scene_id: str, session):
     if not delete_scene(scene_id, session):
         raise ValueError("Scene ID not valid")
 
+    # Remove the scene from scenario's scene_ids array
+    scenarios = get_scenarios(session)
+    for scenario in scenarios:
+        if int(scene_id) in scenario.scene_ids:
+            scenario_obj: Scenario = await get_scenario_from_postgres(
```
jaydhulia, line 216: I am confused... why are you getting the `scenario_obj` here? You already have the scenario (`for scenario in scenarios:`), why is there a need to make a second request to postgres to get the same scenario?

app/library/postgres.py
```diff
@@ -209,6 +209,16 @@ async def delete_scene_from_postgres(scene_id: str, session):
...
+                scenario.id, session
+            )
+            scenario_obj["scene_ids"].remove(int(scene_id))
```
jaydhulia, line 219: `scenario_obj` is not a dict, on line `scenario_obj: Scenario `. So you should use `scenario_obj.scene_ids` if it is a scenario, OR change the above line to  `scenario_dict: dict`. But like the previous comment mentions, scenario_obj is redundant.  
image

jaydhulia (changes requested):

app/library/postgres.py
ahmedhamodi: Because I am being silly. Not needed!

aaronabraham311:
@ahmedhamodi I used the exact same logic that Kevin used [here](https://github.com/uwblueprint/cdc-frontend/blob/69ab646609d3a3a52ae942973617885e37cef58e/src/Components/Admin/EnvironmentEditor/environment.js#L135) for scene creation. Is it worth changing the scene creation endpoint to automatically update the scenario's `scene_ids` field as well?

jaydhulia:
@aaronabraham311 I don't think we should do that. Scenes are independent of the scenario, and scene creation shouldn't automatically add to any scenario. It makes sense for deletion, but not addition imo.

jaydhulia (approved):

PR #102, opened 2021-05-19
Intro conclusion popups

drbalar:
Demo: https://www.loom.com/share/f89205e16c1940e7a9ad2f2947be257b

### DIFF
--- a/app/public/templates/scene.html
+++ b/app/public/templates/scene.html
@@ -165,7 +165,6 @@
             "componentType": "text-pane",
             "jsonData": transition_data
           };
-          console.log(componentDataParsed);
           addEntityToBlackboard(componentDataParsed);
         }
 
--- a/app/public/templates/scene.html
+++ b/app/public/templates/scene.html
@@ -86,7 +86,11 @@
                 position="{{ scene_dict['camera_properties']['position'][0] }} 0 {{ scene_dict['camera_properties']['position'][2] }}">
         <a-entity
           id="primaryCamera"
-          camera
+          {% if cur_scene_idx == 0 %}
+            camera="active: false;"
+          {% else %}
+            camera="active: true;"
+          {% end %}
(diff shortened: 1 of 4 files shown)


scripts/mockdata/scenarios.json
```diff
@@ -54,7 +74,7 @@
         "data": [
           { "text": "just text, no images" },
           {
-            "text": "ANOTHA ONE",
+            "text": "Conclusion: ANOTHA ONE",
```
ahmedhamodi, line 77: Perhaps I'm missing something, but in the video it shows `Conclusion: Just text no image` but here it says `Conclusion: ANOTHA ONE`. Where is the text being rendered from?

ahmedhamodi (commented):
Some lint errors. Otherwise, video looks good from an end user perspective. A small comment on where the actual text is being pulled from, appears to be mismatched?

app/public/templates/scene.html
```diff
@@ -123,37 +127,47 @@
...
+
+          addTransition(transitionData);
+        });
+
+        function addTransition(transition_data) {
+          let componentDataParsed = {
+            "componentType": "text-pane",
+            "jsonData": transition_data
+          };
+          console.log(componentDataParsed);
```
jaydhulia, line 168: `console.log` :D

scripts/mockdata/scenarios.json
drbalar: oh lol the video was from before I did merge with vivian's latest changes

drbalar: great observation 🧐

ahmedhamodi: Ah okay, as long as the functionality is the same then I'm good with that :)

jaydhulia (approved):

PR #103, opened 2021-05-19
Update schema validations to include numbers

ahmedhamodi:
When going through API integration from the frontend, I would be trying to name a room "Escape Room 2" and it would fail due to the lack of allowing numbers. I think numbers should be allowed since this is a very viable name that DCC admins might use.

As a future point, things like apostrophes should also be supported, if I want to do "Billy's Room" or something along those lines.


drbalar (approved):
looks good

PR #105, opened 2021-05-22
Dhruvin intro conclusion

drbalar:
Add intro landing page
Add conclusion completed page

landing page goes to intro popup
conclusion popup goes to completed page which shows a shareable link

demo: https://www.loom.com/share/af99ad28c3384c0ea6c8266a4da42657


drbalar:
Creation of new scenarios still needs to be tested through postman once the cookie setup is worked out. But should be working

ahmedhamodi (commented):
The flow looks great!

@aaronabraham311 - do we want to align the intro/conclusion pages to current mockups? Are those close to finalized or do we expect significant changes to it?

aaronabraham311:
gif

Flow looks great! Great work as always.

@ahmedhamodi I don't think it's worth waiting for the designs. It's likely that Jack will make some changes so don't want to double the work. Let's merge this in for now.

app/public/templates/scene.html
```diff
@@ -148,7 +148,7 @@
               // it is last scene, navigate to same scene for now
               // TODO: in the future, this will redirect to scenario complete screen
```
jaydhulia, line 149: We can remove this TODO now 😄

app/routes/ui/completed_page.py
```diff
@@ -0,0 +1,42 @@
...
+                )
+            else:
+                scenario_obj: Scenario = Scenario(**scenario_dict)
+            if not scenario_obj.is_published:
+                raise ValueError("This Scenario is currently not accessible")
+
+            await self.render(
+                "completed_page.html",
+                scenario_name=scenario_obj.name,
+                scenario_friendly_name=scenario_obj.friendly_name,
```
jaydhulia, line 34: `scenario_friendly_name` isn't used in the completion page right?

app/routes/ui/completed_page.py
```diff
@@ -0,0 +1,42 @@
...
+                scenario_share_link=scenario_obj.conclusion_data["share_link"],
```
jaydhulia, line 35: Just so we don't always have to update this as we add conclusion, I recommend instead doing:
`scenario_conclusion_data=scenario_obj.conclusion_data`
And then accessing it that way in completion page HTML:
```
share_link = "{{ scenario_conclusion_data['share_link' }}";
```
If we don't do it this way, then we have to update this handler every time we update conclusion_data, so would rather avoid that!

jaydhulia (commented):
Works great, nice work! Made some minor comments for changes @drbalar!

app/routes/ui/completed_page.py
```diff
@@ -0,0 +1,42 @@
...
+                json=json,
```
jaydhulia, line 36: I believe `json` is also not needed for this.

app/routes/ui/landing_page.py
```diff
@@ -0,0 +1,42 @@
...
+                scenario_obj: Scenario = Scenario(**scenario_dict)
+
+            if not scenario_obj.is_published:
+                raise ValueError("This Scenario is currently not accessible")
+
+            await self.render(
+                "landing_page.html",
+                scenario_name=scenario_obj.name,
+                scenario_friendly_name=scenario_obj.friendly_name,
+                json=json,
```
jaydhulia, line 36: Same as the completion page, I believe json isn't needed here

app/routes/ui/completed_page.py
drbalar: yeah we dont use it. I had left it in case it would have been needed. But ill remove it for now

app/routes/ui/completed_page.py
drbalar: good point

jaydhulia (approved):
Thanks for the changes, looks good to me!

PR #106, opened 2021-05-23
Load inspector JS statically without needing to run inspector

jaydhulia:
- Added inspector js to the backend repo so you can test out and use the inspector without needing to run the forked inspector version


ahmedhamodi (commented):
Should we use a minified version? 54k line file isn't great to have in the repo I think

jaydhulia:
Yeah, I think it should be okay for now. I have left minifying out for now because we need to minify and combine all the JS files when we are ready. For now, until we get closer to a deployment state, I believe this is sufficient.

ahmedhamodi (approved):

PR #107, opened 2021-05-24
Generate pre-signed S3 PUT URLs that the frontend can use to upload to S3

jaydhulia:
This is the recommended AWS approach for uploading to S3: the backend generates the S3 pre-signed PUT URL with the required condition and fields. And then the frontend uses that information to upload from the frontend.

Please see the first answer here: https://stackoverflow.com/questions/54076283/how-to-upload-a-file-to-s3-using-presigned-url-with-react-js for an example of what needs to be done on the frontend side (using axios).

I have tested this locally using curl and it works 😄  Can upload either image or asset, can also overwrite/update if you provide 
 S3 key. This handler will serve as *THE* handler for all uploads to S3.

 Happy to show a demo during the work session if needed.


configs/dev-config.yaml
```diff
@@ -66,3 +66,10 @@ auth:
   min_since_login: 100
 
 frontend_domain: http://localhost:3000
+
+s3:
+  allowed_types:
+    - asset
```
drbalar, line 72: Are these the special types for AWS S3? Or the types of objects we have in the postgres Assets table?

aaronabraham311:
Nice work!

When a user is uploading an asset, we also want them to create a new record in the asset table. Do we need to fire another request for that or can we bundle them into one asset POST request? Maybe this would be for another ticket (because we don't need this behaviour for picture upload) but curious about it

ahmedhamodi:
> When a user is uploading an asset, we also want them to create a new record in the asset table. Do we need to fire another request for that or can we bundle them into one asset POST request? Maybe this would be for another ticket (because we don't need this behaviour for picture upload) but curious about it

I agree, in my opinion we do want that happening within this POST. However, we can do that later unless Jay is feeling up for it for this particular PR. At the very least, let's put a "TODO" within the type == asset if statement so we know where it'll need to be added.

jaydhulia:
No, this handler won't include (now or in the future) creating a new record in the asset table, so that will need to be a separate POST request once FE uploads the file to S3. We already have a route for handling that (https://github.com/uwblueprint/cdc-backend/blob/master/app/routes/admin/asset.py#L14) and there's no point in duplicating the validation and creation logic from there to here. If we have that here, we would have to add conditional validation on the JSON body of this request, and this handler could throw errors that are totally unrelated to AWS. Furthermore, if we add the record in asset table here, we would be adding an `s3_key` in asset table that is currently not valid. So if the upload fails for some reason, we would have to add logic for deleting the asset. If we added logic for POST function of the asset, then we would also have to handle PUT of the asset table here because this function also allows for updating a file in S3. 
This handler's function is supposed to be completely agnostic to Postgres and other logic (like asset creation/updating). This handler's function is simple: you want to upload something to S3, this function will serve as a proxy so FE can upload to S3 without needing AWS credentials.

ahmedhamodi (approved):

PR #109, opened 2021-05-26
One Room Escape Room

vivianliu0:
Classroom escape room with a simple flow, assets will be added later once they are finalized.

Loom demo: 
https://www.loom.com/share/20a2ce6b116242cc99304998936a2acc

Ahmed edit:
Latest loom video link (as mentioned in a comment below)
https://www.loom.com/share/2467af27d6ba4700940cb17c9d7d54d6


ahmedhamodi (approved):
Video looks great to me :) Great work 🎉

vivianliu0:
Forgot to add the placeholder teacher and student lol

New demo: 
https://www.loom.com/share/0562d431699b428daac45d56fc299a5a

ahmedhamodi (dismissed):

aaronabraham311:
Please do not merge yet. @fakesquid will be uploading a couple of 3D assets that you can use. Once you integrate that, Jack will also need to do a design QA. Couple of other things we need to consider:
- Can we allow the user to move in-between desks? If that requires removing some desks to make space, I'm all for it.
- Let's make this a little bit more challenging of a room by also including some desks with a math textbook with some random math problem and a history book with a date of some event (choose whichever historical event you want). 
- Is there a reason why the desks are positioned weirdly like that? Why do all the interactive puzzles have to be on a desk that is positioned away from the other desks? Can we try to make it look like one uniform set of desks in a regular class configuration?

jaydhulia:
@aaronabraham311:

> Can we allow the user to move in-between desks? If that requires removing some desks to make space, I'm all for it.

We can, but this would mean I have to open blender and go through the process of creating, saving and updating the nav-mesh on our repo. I am happy to do it, but only if this is an actual asset that will be using (don't want to do this work if we are just going to discard this background)

> Is there a reason why the desks are positioned weirdly like that? Why do all the interactive puzzles have to be on a desk that is positioned away from the other desks? 

There's no reason why the desks are positioned weirdly, we did it that way to make it easier to tell which ones are being modified lol. As for the actual puzzles, once we have the assets we can merge that into the actual room (like once we have a textbook asset, we can place it on the chair for example).

> Can we try to make it look like one uniform set of desks in a regular class configuration?

Similar to my first answer, if we want to put one of our chairs in a regular spot (replace one of the other chairs), this would involve blender + nav-mesh work.

vivianliu0:
Updated the PR with a couple of things:
1. The escape room now has the assets that Jack provided
2. Added in a math textbook and a history textbook on two desks to make the room more challenging and throw players off
3. Updated the scene background to use the new version Jay created, which no longer has the block in the middle of the room
4. At the moment we can't navigate in between desks, however I placed the textbook and students on/behind desks and they are still interactable.

New Demo video showing updated flow:
https://www.loom.com/share/2467af27d6ba4700940cb17c9d7d54d6

aaronabraham311 (commented):
Looks pretty good to me! Great work. 

Is it possible to have text on the password input when you click on the teacher? Something like, "what's the answer to the question the teacher just asked" or something along the lines to provide a little bit more context. If this isn't constructed, don't worry about it; I can cut out a new ticket if that's the case

jaydhulia:
@aaronabraham311 
> Is it possible to have text on the password input when you click on the teacher? Something like, "what's the answer to the question the teacher just asked" or something along the lines to provide a little bit more context. If this isn't constructed, don't worry about it; I can cut out a new ticket if that's the case

Yeah we currently don't have this functionality implemented, so a new ticket would be needed. Should be pretty simple though (less than 30 min work)

app/library/api_schemas.py
```diff
@@ -20,7 +20,10 @@
     "properties": {
         "name": {"type": "string", "pattern": r"^[a-zA-Z0-9 _-]{1,50}$"},
         "friendly_name": {"type": "string", "pattern": r"^[a-zA-Z0-9_-]{1,50}$"},
-        "description": {"type": "string", "pattern": r"^[\?\!\.,a-zA-Z0-9 _-]{,2000}$"},
+        "description": {
```
jaydhulia, line 23: Weird, the two regex are identical, not sure what happened for pre-commit to do this automatically

jaydhulia (commented):
`app/public/static/assets/dev/armchair2.gltf` - this file is not used right? Let's remove that

aaronabraham311 (approved):
Design approval!

jaydhulia (approved):
Approval from Dev side, works great 😄

PR #111, opened 2021-05-26
Scene duplication

drbalar:


ahmedhamodi (dismissed):
lgtm :)

app/library/postgres.py
```diff
@@ -234,12 +234,31 @@ async def duplicate_scenario(scenario_id: str, session):
...
 
 async def duplicate_scene(scene_id: str, session):
-    # TODO: actual duplication in postgres
-    sample_response = {
-        "message": "Duplicated scene with id " + scene_id,
-        "id": scene_id + "3",
-    }
-    return sample_response
+    # Get scene to duplicate
+    curr_scene = get_scene(scene_id, session).as_dict()
```
jaydhulia, line 238: I think this needs error checking before trying to access it as dict:
```
    curr_scene = get_scene(scene_id, session)
    if curr_scene is None:
        raise ValueError("Invalid Scene ID")
```

app/library/postgres.py
```diff
@@ -234,12 +234,31 @@ async def duplicate_scenario(scenario_id: str, session):
...
+    new_scene_id = new_scene["id"]
+
+    # Duplicate the objects and add into the scene
+    for object_id in object_ids:
+        curr_object = get_object(object_id, session).as_dict()
+        try:
+            del curr_object["id"]
+        except KeyError:
+            pass
+        await post_object_to_postgres(new_scene_id, curr_object, session)
```
jaydhulia, line 257: Ah, this method is going to have O(n) calls to update the scene, because on each object being created, it will add to the scene (line 268). Since in this case, we are duplicating the scene, instead of making O(n) calls, let's make O(1) call.
So, either add a parameter to  `post_object_to_postgres` like `post_object_to_postgres(scene_id: str, data: dict, session, add_to_scene=True)` and then add an if statement for the parts related to the scene in that method, OR copy the relevant parts of that method to this method, and just do one `put_scene` with all the new object IDs appended to the scene.

app/library/postgres.py
```diff
@@ -234,12 +234,31 @@ async def duplicate_scenario(scenario_id: str, session):
...
+    curr_scene = get_scene(scene_id, session).as_dict()
+    object_ids = curr_scene["object_ids"]
+    try:
+        del curr_scene["id"]
+        del curr_scene["object_ids"]
+    except KeyError:
+        pass
+
+    # Create new duplicated scene without the objects
+    new_scene = await post_scene_to_postgres(curr_scene, session)
```
jaydhulia, line 247: Just so the newly created scene isn't literally indistinguishable from the current scene on the frontend, perhaps we should append `_copy` to the name? Similar to how filesystems typically duplicate files?
Thoughts @ahmedhamodi @aaronabraham311 ?

jaydhulia (changes requested):
Looks good overall, I think some changes need to be made, then will be good!

app/library/postgres.py
aaronabraham311: I agree. In the case of duplicating a duplicate scene (assuming that the scene name wasn't changed prior), would we just add another _copy to it?

jaydhulia: Yup that would make most sense, since realistically we can't go through all the scenes and see how if another _copy already exists. Alternatively, frontend can ask for a scene name when duplicating. Like how some apps (PyCharm) does it. So they would hit "Duplicate" and you would prompt for a scene name.

app/library/postgres.py
drbalar: good point

app/library/postgres.py
drbalar: If we duplicate `scene 1`, have `scene 1` and `scene 1_copy`. 

If we then duplicate `scene 1` again, we will have 2 `scene 1_copy` named scenes. That is fine or we want some logic to avoid having duplicate scene names?

drbalar:
@aaronabraham311  @jaydhulia @ahmedhamodi  For duplicating scenes, we also duplicate it's objects. For the object duplication.. is it just duplication or we should also update its `next_obj` field? If we are deprecating that field then it is not worth i think but otherwise we might need this.

app/library/postgres.py
jaydhulia: Yeah that’s what I was worried about lol, having the frontend provide a modal to choose a name would be nice, but we can wait for @ahmedhamodi or @aaronabraham311 to respond on how much work that would be. 
Otherwise, we shouldn’t do much magic on backend, just simply append _copy and leave it to them to rename.

jaydhulia:
Yeah we are deprecating that field, so feel free to leave it as is for now. We will get rid of that field in a later ticket (is pretty priority) @drbalar

app/library/postgres.py
```diff
@@ -234,12 +234,39 @@ async def duplicate_scenario(scenario_id: str, session):
...
+    new_scene_id = new_scene["id"]
+    new_object_ids = []
+    # Duplicate the objects and get new IDs
+    for object_id in object_ids:
+        curr_object = get_object(object_id, session).as_dict()
+        try:
+            del curr_object["id"]
+        except KeyError:
+            pass
+        curr_object["name"] += "_copy"
```
jaydhulia, line 261: I think we can leave the object names as is, because the objects will be part of a different scene, so there won't be any confusion between objects of initial scene vs the copied scene.

jaydhulia (approved):
Tried this out, works great! Thanks @drbalar

app/library/postgres.py
ahmedhamodi: Yeah I think what we should do is add _copy and display this name, but in a modal where they can edit and then we'd send a PUT request. I think this PR is good.

PR #113, opened 2021-05-26
Changes that make deployment work

jaydhulia:


configs/dev-config.yaml
```diff
@@ -34,10 +34,6 @@ default_data:
           ],
         "currPosition": 0,
       },
-      {
```
drbalar, line 37: This was needed to deploy or just cutting down on extra lines?

drbalar (approved):
just a question, looks good otherwise

configs/dev-config.yaml
jaydhulia: Wasn’t technically needed for the deploy but it’s a bug that I wanted to fix! Because based on our discussion yesterday, a scenario with 0 scenes should have 0 transitions (so only the intro). And since we use the default for scenario creation, it makes sense to fix it

drbalar: Ohh right that makes sense

ahmedhamodi (approved):
lgtm :)

PR #115, opened 2021-05-29
Asset object cascade deletion

drbalar:
- objects referencing to the asset, it will cascade delete.
- if a scene `background_id` references an asset then an appropriate error is raised


app/library/postgres.py
```diff
@@ -102,9 +102,22 @@ async def get_assets_from_postgres(session):
 
 
 async def delete_asset_from_postgres(asset_id: str, session):
-    if not delete_asset(asset_id, session):
+    if not get_asset(asset_id, session):
         raise ValueError("Asset ID not valid")
 
+    scenes_resp = await get_scenes_from_postgres(session)
+    for scene in scenes_resp["scenes"]:
+        if asset_id == str(scene["background_id"]):
```
jaydhulia, line 110: We should have 2 for loops, one for this and after for the deletion (that's outside this for loop), otherwise we may delete half objects, then error.
Like if scene 1 and 3 have the asset as object, and scene 2 has it as background. Then, scene 1's objects will delete, but scene 3's won't because we errored halfway through (when checking scene 2).

app/library/postgres.py
```diff
@@ -102,9 +102,22 @@ async def get_assets_from_postgres(session):
...
+            raise AssertionError(
+                f"Asset ID {asset_id} is reference as background_id in scene with ID {scene['id']}"
+            )
+        for object_id in scene["object_ids"]:
+            object_data = get_object(object_id, session).as_dict()
```
jaydhulia, line 115: FYI, you don't need `as_dict()` here, you can just do:
```
object_data = get_object(object_id, session)
if asset_id == object_data.asset_id:
 ...
```

app/library/postgres.py
```diff
@@ -102,9 +102,22 @@ async def get_assets_from_postgres(session):
...
+            if asset_id == str(object_data["asset_id"]):
+                await delete_object_in_postgres(scene["id"], object_id, session)
```
jaydhulia, line 117: Rather than doing it this way, can you investigate if we can use CASCADE delete in Postgres? CASCADE delete is usually used for this exact reason: when there is a foreign key referencing a key from another table and that other table's key is deleted, then the foreign key references should also be dropped.
https://kb.objectrocket.com/postgresql/how-to-use-the-postgresql-delete-cascade-1369
https://docs.sqlalchemy.org/en/14/orm/cascades.html
If you explore this and find that this is too much work, then ignore this comment. I recognize this might be too much work since we have to remove the object from scene's objects table too. So you may need to explore this first!

drbalar: Looked into CASCADE delete. It is easy to do it for the objects alone but because we also remove from the scene's object array it does get trickier as you mentioned.

I could think of 2 ways with our current schema:
1. We can leave it as is without cascade
2. We use CASCADE delete to get rid of the objects related to the asset first. And then we loop through the scenes and update it's object list to get rid of objects that no longer exist in the Postgres `objects` table.

With option 2 we do save on making multiple calls to the db for deleting objects because of cascade.
Also, if a `scene.object_ids` list is edited then there may be some objects in the database not being used at all. Option 1 would leave them there, option 2 would delete those.

Do you have any thoughts/suggestions?

jaydhulia: Option 2 sounds good to me, we want to make fewer calls to DB but also make sure that our database stays consistent. You can do a query for get_all_objects after CASCADE delete command, and create a set of the IDs. Then loop through the scenes removing any that have an object_id that isn’t in the set of IDs.

ahmedhamodi:
I think Jay has a good grasp of this, lmk if my input is needed at all :)

app/library/postgres.py
drbalar: Sounds good. I don't think we have a get_all_objects. Do you think it is worth to have a db function to get_all_objects (the whole objects)? Otherwise, I'm thinking to have it just return the object IDs.

jaydhulia: Yeah we don’t need the entire object, just IDs. I’m sure sqlalcheny has a way to only return certain columns rather than entire table, so maybe we can look at that?

aaronabraham311:
One small thing: when you are finished, could you please update the errors list with the new error?

jaydhulia (approved):
Works great, thanks @drbalar!

PR #116, opened 2021-06-02
More AWS <> Local dev sync

jaydhulia:


app/__main__.py
```diff
@@ -39,24 +39,25 @@
...
-from routes.user.loading_screen import UserLoadingScreen
-from routes.user.scenario import UserScenarioHandler
-from routes.user.scene import UserSceneHandler
-from routes.user.solved import UserSolvedHandler
+
+# from routes.user.asset import UserAssetHandler
+# from routes.user.loading_screen import UserLoadingScreen
+# from routes.user.scenario import UserScenarioHandler
+# from routes.user.scene import UserSceneHandler
+# from routes.user.solved import UserSolvedHandler
```
ahmedhamodi, line 47: Intentional commented out lines?

app/__main__.py
```diff
@@ -39,24 +39,25 @@
...
-        (r"/api/user/v1/scenario/([0-9]{1,16})", UserScenarioHandler),
-        (r"/api/user/v1/loading_screen", UserLoadingScreen),  # TODO: remove endpoint
+        # (
+        #     r"/api/user/v1/solved/([0-9]{1,16})",
+        #     UserSolvedHandler,
+        # ),
+        # (r"/api/user/v1/asset/([0-9]{1,16})", UserAssetHandler),
+        # (r"/api/user/v1/scene/([0-9]{1,16})", UserSceneHandler),
+        # (r"/api/user/v1/scenario/([0-9]{1,16})", UserScenarioHandler),
+        # (r"/api/user/v1/loading_screen", UserLoadingScreen),
```
ahmedhamodi, line 60: Same question here

ahmedhamodi (approved):
Looks good, thanks for making this change.

Couple of comments around some commented code, was this intentional?

Otherwise, you deployed this to the prod URL and I tested on there, looks good (minus that initial cache issue I had with jpg images, we do want to test if this happens again after cache runner has run with this new configuration setup).

This is good to merge though and we can deal with those issues when they come up!

app/__main__.py
jaydhulia: Yup for both... these are the old user API routes that aren't used atm, so thought it's best to block them out of deployment. We can uncomment if we find out some are needed (for stats for example).

ahmedhamodi: I see. can also remove, but I'll leave that with you

jaydhulia: Yeah, I will leave removing this + the corresponding handlers to a later timestamp when we are closer to launch 😄

configs/dev-aws-config.yaml
```diff
@@ -0,0 +1,79 @@
+# Configuration options for dev
+tornado:
+  port: 8888
+  debug: true # will reload server on each change to python files
+  xsrf: true # need to make sure frontend is set up to work properly with this setting on
+
+postgres:
+  hostname: localhost
```
drbalar, line 8: This should be some IP address or does the aws deployment also use localhost?

configs/dev-aws-config.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+auth:
+  cookie_name: dcc_auth
+  expiry_hours: 1
+  # TODO: update this list, if needed
+  allowed_domains:
+    - uwblueprint.org
+  min_since_login: 10
+
+frontend_domain: http://localhost:3000
```
drbalar, line 66: Also here, should this be non localhost? Or it stays localhost because it is local to the machine it is deployed on?

drbalar (commented):
Looks good, just had a couple questions

configs/dev-aws-config.yaml
jaydhulia: For the actual AWS EC2 instance, this is indeed the IP address for the database, but because this config is meant to be run locally, the Postgres instance you are developing on is still local.

configs/dev-aws-config.yaml
jaydhulia: Similar to my previous comment, this is https://admin-dev.jaydhulia.com/ on the EC2 instance, but because this config file is meant to be run locally, it needs to be localhost. This is because HTTP <-> HTTPS calls don't work

configs/dev-aws-config.yaml
drbalar: Ahh I see

drbalar (approved):

PR #117, opened 2021-06-02
Added tutorial flow

drbalar:
Loom: https://www.loom.com/share/1ff073230d3e428fa0918f003b6d8e73


app/public/templates/landing_page.html
```diff
@@ -11,7 +11,8 @@ <h1>{{ scenario_name }}</h1>
     </button>
     <script type="text/javascript">
       function playGame() {
-        let nextURL = "/{{ scenario_friendly_name }}/0";
+        let nextURL = "/{{ scenario_friendly_name }}/tutorial";
+        alert("You are entering the tutorial.");
```
ahmedhamodi, line 15: I think we can remove this. Any reason we want to alert? You mentioned in the loom video that you intentionally left it in, but I don't see the use case. Please let me know :)

aaronabraham311 (approved):
Product approval. Your Loom videos are too good 😆 

When we use Vivian's one room escape room, will this tutorial also be integrated in there once we merge this PR?

app/public/templates/landing_page.html
drbalar: I did not know how the modal will look like so I added that alert as a placeholder for the modal.

Should I remove it?

ahmedhamodi: Oh we have a modal prior to going into tutorial? I see. Okay we can leave that then

app/public/templates/tutorial.html
```diff
@@ -0,0 +1,229 @@
...
+      });
+    </script>
+  </head>
+  <body>
+    <div id="scene-root">
+      <a-scene id="3-scene" cursor="rayOrigin:mouse" raycaster="objects: .link">
+        <a-assets timeout="50000">
+          <a-asset-item
+            id="1-asset-src"
+            src="{{ asset_prefix_url }}assets/dev/armchair.glb"
```
jaydhulia, line 43: Once you merge latest master in, can you update these to the new file versions of them?
https://github.com/uwblueprint/cdc-backend/blob/master/scripts/mockdata/assets.json

This commit shows the diff: https://github.com/uwblueprint/cdc-backend/pull/116/commits/e4245aa4297c983224810dedfb7200b14302b63d

jaydhulia (approved):
Works great @drbalar, nice work! I just fixed one of the image links, other than that looks awesome!

PR #118, opened 2021-06-05
Intro conclusion customizing

drbalar:
added fields to the backend scenario to allow customizing of intro and conclusion text.

Intro:
- header text

Conclusion:
- header text
- paragraph text
- share link

All fields are optional.

UI did not really change so won't demo this yet.

### DIFF
--- a/app/library/api_schemas.py
+++ b/app/library/api_schemas.py
@@ -45,10 +45,20 @@ admin_scenario_put_handler_schema = {
         "publish_link": {"type": "string", "pattern": r"^[\S]{1,50}$"},
         "preview_link": {"type": "string", "pattern": r"^[\S]{1,50}$"},
         "expected_solve_time": {"type": "string", "pattern": r"^[a-zA-Z0-9 _-]{,50}$"},
+        "introduction_data": {
+            "type": "object",
+            "properties": {
+                "header_text": {"type": "string"},
+            },
+            "additionalProperties": False,
+        },
         "conclusion_data": {
             "type": "object",
-            "properties": {"share_link": {"type": "string"}},
-            "required": ["share_link"],
+            "properties": {
+                "header_text": {"type": "string"},
+                "paragraph_text": {"type": "string"},
+                "share_link": {"type": "string"},
+            },
             "additionalProperties": False,
         },
         "transitions": {
--- a/app/models/scenario.py
+++ b/app/models/scenario.py
@@ -16,7 +16,18 @@ class Scenario(BaseModel):
     publish_link = Column(Text, nullable=True)
     preview_link = Column(Text, nullable=True)
     expected_solve_time = Column(Text, nullable=False, default="")
-    conclusion_data = Column(JSONB, nullable=False, default={"share_link": ""})
+    introduction_data = Column(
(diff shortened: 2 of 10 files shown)


app/library/api_schemas.py
```diff
@@ -45,10 +45,20 @@
         "publish_link": {"type": "string", "pattern": r"^[\S]{1,50}$"},
         "preview_link": {"type": "string", "pattern": r"^[\S]{1,50}$"},
         "expected_solve_time": {"type": "string", "pattern": r"^[a-zA-Z0-9 _-]{,50}$"},
+        "introduction_data": {
+            "type": "object",
+            "properties": {
+                "header_text": {"type": "string"},
```
ahmedhamodi, line 51: Do we also want a paragraph text for intro? Or just conclusion?

app/routes/ui/landing_page.py
```diff
@@ -31,9 +31,11 @@ async def get(self, scenario_friendly_name):
                 "landing_page.html",
                 scenario_name=scenario_obj.name,
                 scenario_friendly_name=scenario_obj.friendly_name,
+                scenario_introduction_data=scenario_obj.introduction_data,
             )
         except ValueError as e:
             self.write_error(status_code=404, message=str(e))
-        except Exception:
+        except Exception as e:
+            print(e)
```
ahmedhamodi, line 39: Do we want this print still?

ahmedhamodi (commented):
Some thoughts, but overall code lgtm

app/routes/ui/landing_page.py
drbalar: Yikes, no

app/library/api_schemas.py
drbalar: I imagined the intro being just a title and a play button. What do you think?

ahmedhamodi: I'm good with this for now, but we may want to add in the future.

app/models/scenario.py
```diff
@@ -16,7 +16,18 @@ class Scenario(BaseModel):
     publish_link = Column(Text, nullable=True)
     preview_link = Column(Text, nullable=True)
     expected_solve_time = Column(Text, nullable=False, default="")
-    conclusion_data = Column(JSONB, nullable=False, default={"share_link": ""})
+    introduction_data = Column(
+        JSONB, nullable=False, default={"header_text": "Welcome to the escape room."}
```
jaydhulia, line 20: Let's make this configurable, similar to how we do transitions, so we can change it easily once DCC gives us the default ones they want. So like:
```
default=config.get("default_data.introduction_data")
```

app/models/scenario.py
```diff
@@ -16,7 +16,18 @@ class Scenario(BaseModel):
...
+    )
+    conclusion_data = Column(
+        JSONB,
+        nullable=False,
+        default={
```
jaydhulia, line 25: Let's make this configurable, similar to how we do transitions, so we can change it easily once DCC gives us the default ones they want. So like:
```
default=config.get("default_data.conclusion_data")
```

jaydhulia (approved):
Couple of minor comments, LGTM otherwise! Approved!

jaydhulia (approved):

PR #119, opened 2021-06-05
Backend changes for screenshot for scenes

jaydhulia:
- [x] Attach default generic image to scene if no screenshot exists
- [x] If screenshot key exists, attach the appropriate domain (backend domain, or S3 depending on where we are working)
- [x] On deletion, delete from AWS if the scene has a screenshot and env has `aws`
- [x] On duplication, duplicate the image in S3 if we are working in env `aws`


app/library/postgres.py
```diff
@@ -254,6 +280,7 @@ async def delete_scene_from_postgres(scene_id: str, session):
                 await update_scenario_from_postgres(
                     scenario.id, scenario.as_dict(), session
                 )
+                break
```
drbalar, line 283: is this because a scenario will only refer to a scene once?

drbalar (approved):
Code looks good 🎉, just a question i had

app/library/postgres.py
jaydhulia: Yup exactly, no need to go through the rest of the scenario's scene ids!

drbalar (approved):

configs/dev-config.yaml
```diff
@@ -2,7 +2,7 @@
 tornado:
   port: 8888
   debug: true # will reload server on each change to python files
-  xsrf: true # need to make sure frontend is set up to work properly with this setting on
+  xsrf: false # can keep this false for dev mode
```
ahmedhamodi, line 5: Any reason we need to change this? I think frontend is working with xsrf tokens now?

jaydhulia: Yeah that's why I am changing it to false by default, when testing with FE auth stuff, this can be set to true. I am changing it to false because for Postman, it requires too many extra unnecessary steps.

ahmedhamodi: So is the plan to deal with that issue in the future? Like change this back to xsrf: true and deal with those extra unnecessary steps?

jaydhulia: Not sure what you mean by deal with the issue. It’s not an issue, it is what is expected when using Postman because it’s not a browser. Those extra steps will always be needed if using postman and having this set as true.

ahmedhamodi: Hmm I see. Maybe I'm misunderstanding, but I trust that this is good to go. Just wanted to ask!

PR #121, opened 2021-06-09
WIP: Customize keyboard text

drbalar:
Via the backend, the keyboard/keypad text can now be customized. The font size does not seem changeable, just the text and its color.

image

### DIFF
--- a/app/public/static/js/components/keypad.js
+++ b/app/public/static/js/components/keypad.js
@@ -92,7 +92,6 @@ AFRAME.registerComponent("keypad", {
       this.solvedPuzzleEntity.setAttribute("visible", "true");
     } else {
       // Not solved yet
-      console.log(data);
       keypad_label = data.hasOwnProperty("keypad_text")
         ? data.keypad_text
         : "Enter Password";
--- a/app/public/static/js/components/keypad.js
+++ b/app/public/static/js/components/keypad.js
@@ -92,12 +92,15 @@ AFRAME.registerComponent("keypad", {
       this.solvedPuzzleEntity.setAttribute("visible", "true");
     } else {
       // Not solved yet
-      keypad_label = data.hasOwnProperty("keypad_text")
+      keypad_text = data.hasOwnProperty("keypad_text")
         ? data.keypad_text
         : "Enter Password";
+      keypad_text_color = data.hasOwnProperty("keypad_text_color")
+        ? data.keypad_text_color
+        : "white";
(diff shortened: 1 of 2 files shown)


app/public/static/js/components/keypad.js
```diff
@@ -91,18 +92,25 @@ AFRAME.registerComponent("keypad", {
       this.solvedPuzzleEntity.setAttribute("visible", "true");
     } else {
       // Not solved yet
+      keypad_label = data.hasOwnProperty("keypad_text")
+        ? data.keypad_text
+        : "Enter Password";
       el.setAttribute("super-keyboard", {
-        label: "Enter Password",
+        label: keypad_label,
         labelColor: "black",
```
ahmedhamodi, line 100: Can we make this customizable? Or at the very least default to a non-black colour, it's a bit hard to see imo

drbalar: Made the default `white` and also customizable

aaronabraham311:
Overall looks good! Is the text size not changeable throughout the entire application?

drbalar:
> Overall looks good! Is the text size not changeable throughout the entire application?

No, i cant find `font-size` in the aframe docs. With the normal text we have to scale it up or down. For the keyboard's text, we have to scale the whole keyboard component since the text is a part of it.

app/public/static/js/components/keypad.js
```diff
@@ -91,19 +92,32 @@ AFRAME.registerComponent("keypad", {
...
-        labelColor: "black",
+        label: keypad_text,
+        labelColor: keypad_text_color,
       });
     }
   },
 });
 
 async function removeError(el) {
+  const data = this.data.jsonData;
```
jaydhulia, line 110: Does `this` work here? Isn't `this` not defined? You may instead want to pass in the data in the method where we call `removeError`

drbalar: good catch

jaydhulia:
`Is the text size not changeable throughout the entire application?`
@drbalar you can look at this: https://aframe.io/docs/1.1.0/components/text.html#sizing and https://aframe.io/docs/1.1.0/components/text.html#auto-scaling to look at text size and changing it. I recall changing it when I was playing around with it

drbalar:
Thats a good find. Talked with Jay, Going to update a `title/header` text for all blackboard components. This way should be more consistent. Will create a new pr for it

PR #122, opened 2021-06-09
Take and upload scene screenshot to S3 for a scene

jaydhulia:
Saves a screenshot of the scene on initial render of admin pages to S3 and updates the screenshot key in Postgres

Before first load of history classroom:
image

After the first load of history classroom:
image


aaronabraham311 (commented):
Couple of things I wanted to confirm:
1. Do we overwrite the screenshot upon a new background ID being selected?
2. Do we duplicate the image upon scene duplication?
3. Is the endpoint to access a particular scene updated with the screenshot URL?

jaydhulia:
>Do we overwrite the screenshot upon a new background ID being selected?

This is going to require changes to the inspector, so it will be on PR in the inspector repo. 

>Do we duplicate the image upon scene duplication?

Yes, it duplicates on S3, and deletes from S3 if scene is deleted.

> Is the endpoint to access a particular scene updated with the screenshot URL? If you could demonstrate this with a quick Loom demo, that would be best

Just added a screenshot demonstrating functionality

app/public/templates/scene.html
```diff
@@ -178,9 +181,72 @@
...
+          xhr.send(JSON.stringify(requestData));
+      }
+
+      function dataURItoBlob(dataURI) {
+          const binary = atob(dataURI.split(',')[1]);
+          const array = [];
+          for(let i = 0; i < binary.length; i++) {
+                array.push(binary.charCodeAt(i));
+            }
+            return new Blob([new Uint8Array(array)], {type: 'image/jpeg'});
```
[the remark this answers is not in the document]
jaydhulia: Yes, it will be

aaronabraham311 (approved):
🚢

app/public/static/js/libraries/aframe-inspector.js
```diff
@@ -29894,12 +29894,12 @@ object-assign
...
             key: "getRequests",
             value: function getRequests(self) {
-              var baseUrl = "https://backend-dev.jaydhulia.com/";
+              var baseUrl = "http://localhost:8888/";
               var apiEndpointScene = AFRAME.scenes[0]
                 .getAttribute("id")
                 .replace("-scene", "");
               var baseEndpoint = "api/admin/v1/";
-              var assetsUrl = "https://backend-dev.jaydhulia.com/static/";
+              var assetsUrl = "http://localhost:8888/static/";
```
drbalar, line 29902: this is because the inspector and backend run "locally" on EC2?

app/routes/admin/scene.py
```diff
@@ -118,3 +119,27 @@ async def get(self):
 
         except Exception as e:
             self.write_error(status_code=500, message=str(e))
+
+
+class AdminSceneScreenshotHandler(BaseAdminAPIHandler):
```
drbalar, line 124: im unaware of some context.. this one is for manually invoking a screenshot and update?

drbalar (approved):
@jaydhulia  just a couple questions

Could not test because aws but codewise looks good!

app/public/static/js/libraries/aframe-inspector.js
jaydhulia: It's for running locally since the purpose of this inspector static file is to make it work locally without having to run inspector, so it needs to talk to localhost backend, rather than https backend

app/routes/admin/scene.py
jaydhulia: This one is for updating the screenshot URL link, so we don't have to get entire scene JSON and send updated version for updating just 1 field (screenshot_url)

PR #123, opened 2021-06-10
Text header for blackboard.

drbalar:
For any blackboard object, this will work. Currently text and color can be modified. Color is white by default.
The text will show on the top of the puzzles. It will auto resize based on the amount of text and wrap around if needed. About 25-30 words is the limit right now before we overflow off the cameraview.

Below are different scenarios of the blackboard puzzles. 

image
image
image
image

### DIFF
--- a/app/public/static/js/components/animation-on-click.js
+++ b/app/public/static/js/components/animation-on-click.js
@@ -35,8 +35,8 @@ AFRAME.registerComponent("animation-on-click-run", {
 
 // TODO: Add support for multiple entity creation
 function addEntityToBlackboard(componentDataParsed) {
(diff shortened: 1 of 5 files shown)


drbalar:
Also for keypad puzzle, SOLVED and ERROR also use the same text. 

image
image

jaydhulia (approved):
Nice work @drbalar, works great and looks good to me 🎉

ahmedhamodi:
Love the detailed images showing the text. Looks good to me from a end-user perspective :)

ahmedhamodi (approved):

aaronabraham311:
Thank you Dhruvin, really nice work!

PR #124, opened 2021-06-12
Different mouse/cursor modes

drbalar:
Added 3 cursor modes.

Based on the query param, switch the cursor mode for interaction. The following modes are done:

- The normal mouse cursor we have right now. Mouse click happens where you click the mouse. (default) (http://localhost:8888/teacher-escape-room/0?cursor_mode=1)
- A crosshair that always stays centered. You click once and enter pan mode. Kinda like corona escape. (http://localhost:8888/teacher-escape-room/0?cursor_mode=2)
- A crosshair that always stays centered. Click and drag to pan around. (http://localhost:8888/teacher-escape-room/0?cursor_mode=3)

With crosshair looks like this right now:
image


app/public/templates/scene.html
```diff
@@ -93,10 +101,29 @@
           {% end %}
           fov="80"
           {% if scene_dict["camera_properties"]["look_controls"] %}
-            look-controls
+            {% if cursor_mode == 2 %}
+              {% if cur_scene_idx == 0 %}
```
ahmedhamodi, line 105: Why do we need to do this for the first scene? Reason I ask is that we don't do this check in the old code, so just wondering what edge case this is testing.

drbalar: There is some bug (maybe in aframe or look-controls). Without this if i just put `look-controls:...` then it hides the mouse cursor in the intro popup even though it should not hide it.

ahmedhamodi: Huh strange. @jaydhulia any thoughts?

aaronabraham311:
My opinion is that we don't merge it yet until our team has a chance to test it. I can schedule 5-10 minutes of testing next session where the entire team can play around with all three mechanics

app/routes/ui/scenario.py
```diff
@@ -63,6 +63,7 @@ async def get(self, scenario_friendly_name, scene_number=0):
                 + "-navmesh"
                 + ".gltf"
             )
+            cursor_mode = int(self.get_argument("cursor_mode", "1"))
```
jaydhulia, line 66: We should catch the `ValueError` exception here and throw a generic exception message here, because otherwise whatever the user writes will be reflected back in HTML, which isn't good security practice usually.
Example: If you go to http://localhost:8888/teacher-escape-room/0?cursor_mode=abc, you will see `abc` reflected in the HTML page

app/routes/ui/scenario.py
```diff
@@ -77,6 +78,7 @@ async def get(self, scenario_friendly_name, scene_number=0):
                 cur_scene_idx=scene_number_int,
                 scenario_transitions=scenario_obj.transitions,
                 take_screenshot=False,
+                cursor_mode=cursor_mode,
```
jaydhulia, line 81: We lose this information if we complete the scene, so we need to update scene.html's `nextURL` to be something like:
```
nextURL = "/{{ scenario_friendly_name }}/" + next_scene_idx + "?cursor_mode=" + {{ cursor_mode }};
```

app/public/templates/scene.html
```diff
@@ -123,7 +150,7 @@
           scale="{{ object['scale'][0] }} {{ object['scale'][1] }} {{ object['scale'][2] }}"
           rotation="{{ object['rotation'][0] }} {{ object['rotation'][1] }} {{ object['rotation'][2] }}"
           {% if object["is_interactable"] %}
-            class="link"
+            class="mainLink"
```
jaydhulia, line 153: Could you clarify why we need mainLink and link now?

app/public/templates/scene.html
```diff
@@ -93,10 +101,29 @@
...
+              {% else %}
+                look-controls="enabled: true; pointerLockEnabled: true;"
+              {% end %}
+            {% else %}
+              look-controls
+            {% end %}
           {% end %}
           position="0 {{ scene_dict['camera_properties']['position'][1] }} 0"
         >
+        {% if not cursor_mode == 1 %}
```
jaydhulia, line 116: I assume this can be simplified to cursor_mode == 2 or cursor_mode == 3 , since we only support 3 cursor_modes in total?

app/public/templates/scene.html
jaydhulia: Explored this and basically, it is "expected" behaviour. When you have multiple cameras, one even if you switch the active camera, what you do in the other camera affects the first camera, unless you remove the relevant properties.
For example, if you go to our current deployment, open any puzzle and press any directional keys, once you exit out, you will see the room camera actually moved. 
This can be fixed by removing the appropriate property when switching camera and then adding it back, but this is super low priority bug I would say. So we should be okay doing it this way for now.

jaydhulia (commented):
@drbalar I made some comments on this PR. Note that because we aren't going to be landing this (we aren't going to support all cursor modes), you don't need to work on them. Once we decide on a cursor mode to implement fully, please implement the relevant comment for that particular cursor_mode.

jaydhulia:
Forgot to mention one thing @drbalar, I noticed that for hovering over interactable objects vs non-interactable objects, it is the mouse pointer that changes from a regular pointer to a "hand" pointer. For this distinction, is it possible to make the cursor itself change? Like cursor crosshair is white for non-interactable objects, but color changes to blue when moving over interactable objects for example.

drbalar:
We have decided to go with mode 1 approach. It is the same implementation we have right now this can be closed.

PR #125, opened 2021-06-13
Modify screenshots on change of background ID

jaydhulia:


app/public/templates/scene.html
```diff
@@ -190,22 +195,26 @@
           {% end %}
           inspector.openInspector();
       });
-      async function takeSceneScreenshot(){
-          const screenshot_canvas = scene.components.screenshot.getCanvas('perspective');
-          await uploadScreenshotToS3(screenshot_canvas.toDataURL("image/jpeg"));
+      async function takeSceneScreenshot(s3_key=""){
+        if(initial_load_js || (!initial_load_js && s3_key !== "")){
```
drbalar, line 199: Minor thing, can we make this into `if(initial_load_js || s3_key !== ""){`?
Unless you kept it like this intentionally for readability

drbalar (approved):
looks good, just a comment

app/public/templates/scene.html
ahmedhamodi: Agreed. From an outside perspective, I think `initial_load_js || (!initial_load_js && s3_key !== "")` is less readable than what Dhruvin suggested here. And yeah, appears logically equivalent to me.

app/public/templates/scene.html
```diff
@@ -237,6 +248,10 @@
           };
           xhr.send(JSON.stringify(requestData));
       }
+      function getCookieScene(name) {
+          let r = document.cookie.match("\\b" + name + "=([^;]*)\\b");
+          return r ? r[1] : undefined;
+        }
```
ahmedhamodi, line 254: The nesting of function and the closing } aren't the same ( the } appears to be 2 spaces deeper ).

drbalar (approved):
Looks good 🎉 🚀

PR #127, opened 2021-06-14
Updated tutorial popup html to match with the rest of the code.

drbalar:


ahmedhamodi (commented):
Does that entity just go "over" the text from the previous pane? I wonder if there is a better fix that actually removes the previous text (or maybe I'm not understanding this change)

jaydhulia (approved):
Looks good to me!

drbalar:
> Does that entity just go "over" the text from the previous pane? I wonder if there is a better fix that actually removes the previous text (or maybe I'm not understanding this change)

I forgot to add the text element to the tutorial's html page but the javascript code expects there to be an element called `blackboadTextEl`. So it was running into an `attribute not found` error and throwing before it can clear the elements in the `wipeBlackboard()` function.

PR #131, opened 2021-06-20
Sample escape room - Scene 2

jaydhulia:
Don't review yet


app/routes/ui/admin_scene.py
```diff
@@ -39,7 +39,7 @@ async def get(self, scene_id):
                     take_screenshot = True
 
             await self.render(
-                "scene.html",
+                "scene_nav_mesh.html",
```
jaydhulia, line 42: TODO: a reminder for myself, change this back to scene.html

drbalar:
What is left TODO:
- [ ] find a non-animated man or figure out how to take out animations. The current man is a walking animation and causes a lot of stuttering.
- [ ] Check if may26 on frame is readable or needs to change
- [ ] Lighting needed?
- [ ] Get unordered puzzle agenda pictures
- [ ] Edit nav-mesh to prevent going through door (and persons too maybe?)
- [ ] Add roof/ceiling to the room

drbalar:
Loom demo: https://www.loom.com/share/1abb0875313a4851889a1ecc1aac4966

drbalar:
@jaydhulia cant make you reviewer for your own PR so tagging you here instead

app/public/templates/scene_nav_mesh.html
```diff
@@ -13,7 +13,7 @@
       <a-scene
         cursor="rayOrigin:mouse"
         raycaster="objects: .link"
-        inspector-plugin-recast
+        inspector-plugin-recast="serviceURL: http://localhost:3000;"
```
ahmedhamodi, line 16: Should we use a variable and define it somewhere that will be easy to change later?

ahmedhamodi (commented):
Small comment, only thing that seemed odd to me. Rest of PR seems fine

app/public/templates/scene_nav_mesh.html
jaydhulia: This one is just a file that we use to generate the navmesh that won't be exposed or rendered in our actual application. So it is okay (rather I would prefer) to leave it as is, rather than putting it in a config variable.

aaronabraham311 (approved):
Great work, Dhruvin and Jay! Approving for product/design

jaydhulia:
Can't approve this since i opened this, looks great to me! Let's 🚢

PR #132, opened 2021-06-21
Allow single quotes in visual pane, hints etc.

jaydhulia:


app/public/templates/scene.html
```diff
@@ -125,12 +125,17 @@
...
             class="link"
             animation__onclick="property: position; dur: 0; startEvents: click"
-            animation-on-click-run='blackboardData: {% raw json.dumps(object["animations_json"]["blackboardData"]) %}'
+            animation-on-click-run='blackboardData: {{ json_encode(object["animations_json"]["blackboardData"]) }}'
           {% end %}
         ></a-gltf-model>
         {% end %}
       </a-scene>
       <script>
+        function htmlDecode(input) {
```
ahmedhamodi: I know I wasn't set as a reviewer, but this function isn't being used? Also a console log

ahmedhamodi: Oh you read my mind

jaydhulia: Already removed it

app/public/templates/scene.html
```diff
@@ -139,7 +139,7 @@
           const next_scene_idx = {{ cur_scene_idx }} + 1;
 
           if({{cur_scene_idx }} == 0){
-            const intro_transition_data = {% raw json.dumps(scenario_transitions[0]) %};
+            const intro_transition_data = {% raw json_encode(scenario_transitions[0]) %};
```
ahmedhamodi, line 142: In the above two cases, you removed the `% raw` but here (and below) it is kept. Any reason?

jaydhulia: Yeah, those 2 are in HTML context so we need it to be escaped so HTML will read it properly. In these later ones, we are in the "script" tag, so it's interpreted as JS, so need to have it as is (hence raw)

ahmedhamodi (approved):
lgtm :)

drbalar (approved):
Looks good 🎉

PR #133, opened 2021-06-23
Allow apostrophes in name/description fields

ahmedhamodi:
In the mock data (as per #132), a single quote was added to the description field of a room (which makes sense).

However, the handlers for this do not accept this, so when I was changing the order of rooms on the frontend, I would get an error.

Updated the validation to allow for apostrophes.


jaydhulia (approved):
LGTM!

PR #136, opened 2021-06-30
Scales the objects to 1.04x their current scale when mouse hovers over.

drbalar:
Loom: https://loom.com/share/278d3f9c14c5442eaf724b90f12c995d

One thing I noticed is that larger objects like the door and people are much more noticeable because they are so large ofc. And similarly, smaller objects like the ones on the bookshelf don't grow too much in the camera's perspective. 

So from testing I found this to a good value. But we can change if we prefer another size.


ahmedhamodi (commented):
Mentioned to Dhruvin directly but I think we should look into a pixel increase instead of 4%.

Something like `oldScale * (oldHeight + 20)/oldHeight` on all three dimensions (height, width, depth) might work better?

drbalar:
Updated loom: https://www.loom.com/share/23a985bf2e124301b10b5d52b94ce918

ahmedhamodi:
Yeahhh this looks a lot better!!

app/public/static/js/components/interactable.js
```diff
@@ -0,0 +1,42 @@
+AFRAME.registerComponent("interactable", {
+  schema: {
```
jaydhulia, line 2: Do you ever use these schema parameters? `axis` or `value`? Are they leftover code from before?

app/public/static/js/components/interactable.js
```diff
@@ -0,0 +1,42 @@
...
+      const z = size.z;
+      let scaling_factor = Math.max(x, y, z);
+      scaling_factor = (scaling_factor + 0.05) / scaling_factor;
+      const new_scale_x = old_scale_x * scaling_factor;
+      const new_scale_y = old_scale_y * scaling_factor;
+      const new_scale_z = old_scale_z * scaling_factor;
+
+      el.setAttribute(
+        "animation__mouseenter",
+        `property: scale; to: ${new_scale_x} ${new_scale_y} ${new_scale_z}; dur: 100; startEvents: mouseenter`
```
jaydhulia, line 34: Do we want the animation to repeat forever while the user hovers over it, kinda like pulsating or just the once scale up on mouse enter? @aaronabraham311

jaydhulia (approved):
One minor comment that may need cleanup, but otherwise from a code perspective this looks good to me!

app/public/static/js/components/interactable.js
aaronabraham311: No just the one scale up like what we currently have

app/public/static/js/components/interactable.js
drbalar: it was leftover lol

ahmedhamodi (approved):

PR #139, opened 2021-07-03
Rework Aframe Modal

drbalar:
https://www.loom.com/share/a7aeb0f4cb0e4330a433b3bdb90ccb38

### DIFF
--- a/app/public/templates/tutorial.html
+++ b/app/public/templates/tutorial.html
@@ -108,9 +108,16 @@ controls: checkpoint, gamepad, trackpad, keyboard, touch;"
           </a-entity>
           <a-entity
             id="returnButton"
-            close-popup='jsonData: {"width": "1", "height": "1", "depth": "0.01", "color": "red", "x": "-8.25", "y": "6.75", "z": "0.005"};'
+            close-popup='jsonData: {"width": "1", "height": "1", "depth": "0.01", "color": "red", "opacity": "0", "x": "-8.25", "y": "6.75", "z": "0.005"};'
             class="link"
-          ></a-entity>
+          >
+            <a-entity
+              id="xClose"
+              text="align: center; width: 1; wrapCount: 1; value: X"
+              position="0 0 0"
+            ></a-entity>
+          </a-entity>
+
           <a-entity
             id="blackboardText"
             text="align: center; width: 15;"
--- a/app/public/templates/tutorial.html
+++ b/app/public/templates/tutorial.html
@@ -117,6 +117,11 @@ controls: checkpoint, gamepad, trackpad, keyboard, touch;"
             position="0 6 0.005"
           ></a-entity>
           <a-entity
+            id="blackboardParagraph"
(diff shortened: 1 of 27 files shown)


ahmedhamodi:
I think the change from yellow target squares to that grey was a good choice. I like it :)

aaronabraham311 (commented):
Nice work Dhruvin! Two things:
1. Can we change the red button to exit the modal to a white X? This would conform with the rest of the design
2. Is it possible to constrain the puzzle pieces for unordered, ordered and jigsaw to not overlap with the text in the pane?

drbalar:
> Nice work Dhruvin! Two things:
> 
> 1. Can we change the red button to exit the modal to a white X? This would conform with the rest of the design
> 2. Is it possible to constrain the puzzle pieces for unordered, ordered and jigsaw to not overlap with the text in the pane?

Yes both should be possible, will make changes in a bit

drbalar:
for new changes: https://www.loom.com/share/f3175d3fb16c4b82bd62a6d0f05b02b1

drbalar:
new link. other one was broken: https://www.loom.com/share/8acfd3186bee498497af7edc6e9d179e

aaronabraham311 (commented):
Nice, looks better. Can we make the X a little smaller? It dominates that corner a little, but we want it a teeny bit more subtle

jaydhulia (dismissed):
Works great, nice @drbalar!

app/public/templates/tutorial.html
```diff
@@ -103,19 +103,31 @@
...
-            close-popup='jsonData: {"width": "1", "height": "1", "depth": "0.01", "color": "red", "x": "-8.25", "y": "6.75", "z": "0.005"};'
+            close-popup='jsonData: {"width": "1", "height": "1", "depth": "0.01", "color": "red", "opacity": "0", "x": "-8.25", "y": "6.75", "z": "0.005"};'
             class="link"
-          ></a-entity>
+          >
+            <a-entity
+              id="xClose"
+              text="align: center; width: 1; wrapCount: 1; value: X"
+              position="0 0 0"
+            ></a-entity>
```
ahmedhamodi, line 118: Correct me if I'm wrong, but I think this is the logic where the "X" button gets rendered in the top right Dhruvin? So in regards to Aaron's comment, I don't know how you can really scale it down. Two ideas (but not sure if they'll work):
- change the X to x (ie. lower case). If they preserve that on render, it'll be smaller by nature 
- change the width attribute (and maybe add a height attribute if it exists) to make it smaller by constraint

Let me know if that works or not!

drbalar:
image
@aaronabraham311 is this a good size?

jaydhulia (approved):

app/public/templates/tutorial.html
drbalar: Tried a few things. I think keeping it big X and just decreasing the width works the best here since it also keeps it centered within its parent element

PR #141, opened 2021-07-07
Asset Screenshot

vivianliu0:
- Added new /api/admin/v1/asset/{id)/screenshot endpoint
- Updated asset table to include screenshot_url
- Updated mock data
- Allow for hard deletes


configs/dev-aws-config.yaml
```diff
@@ -20,7 +20,7 @@ asset:
   prefix_url: https://dcc-bp-public-dev.s3.us-east-2.amazonaws.com/
   # This will be changed to true for prod... for now until we get asset CRUD working
   # don't actually delete from S3
-  aws_hard_delete: false
+  aws_hard_delete: true
```
jaydhulia, line 23: Ah let's keep this false, so we don't accidentally delete our .glb files :)

app/public/templates/scene.html
```diff
@@ -267,4 +266,4 @@
     {% end %}
   </script>
   </body>
-</html>
+</html>
```
jaydhulia, line 269: Missing newline?

app/library/postgres.py
```diff
@@ -149,6 +149,13 @@ async def delete_asset_from_postgres(asset_id: str, session):
             Key=asset_obj.s3_key,
         )
 
+        pos_key = asset_obj.screenshot_url.find("image/") + 6
```
jaydhulia, line 152: Ah, the `image/` is actually part of the key in S3. So the actual key for objects would be like `image/abcd.png`. So you don't need to do + 6 here

app/library/postgres.py
```diff
@@ -149,6 +149,13 @@ async def delete_asset_from_postgres(asset_id: str, session):
             Key=asset_obj.s3_key,
         )
 
+        pos_key = asset_obj.screenshot_url.find("image/")
```
jaydhulia, line 152: Ah, forgot one thing, some images won't have screenshot URL, so this will be an error. I would suggest nesting the rest of this under an if statement like `if pos_key != -1`

vivianliu0: ooo thanks for catching that, updated

jaydhulia (approved):
LGTM!

PR #145, opened 2021-07-10
Jigsaw endpoint

jaydhulia:
Given a base64 encoded image, this endpoint will divide the image into rows x columns (default 3 x 3), and upload all the images to S3. In response, it will return the uploaded image URLs so that they can be associated with the puzzle.

Sample POST body:
```json
{ "encoded_image": "iVBORw0KGgoAAAANSUhEUgAAALkAAADVCAYAAAAL...."}
```

Sample response:
```json
{
  "data": [
    "url1",
    "url2",
     "...."
   ]
}
```


ahmedhamodi:
@amoliksingh as a good test for this (since you are gonna be implementing jigsaw this week) is to run your backend on this branch and use it in the frontend context. See if it works as you expect.

I'll also do some testing individually but I think you will be able to test it best.

ahmedhamodi (approved):
Code looks good to me. @amoliksingh verified it was working when using the endpoint on the admin side, so I think this is good to go. Good work :)

PR #146, opened 2021-07-16
Add pan modes for tutorial page

jaydhulia:


aaronabraham311 (approved):
Thank you!

app/routes/ui/tutorial.py
```diff
@@ -27,11 +27,18 @@ async def get(self, scenario_friendly_name):
...
+                pan_mode = int(self.get_argument("pan_mode", "0"))
+            except ValueError:
+                pass
+
             await self.render(
                 "tutorial.html",
                 asset_prefix_url=config.get("asset.prefix_url"),
                 scenario_name=scenario_obj.name,
                 scenario_friendly_name=scenario_obj.friendly_name,
+                pan_mode="true" if pan_mode == 1 else "false",
```
ahmedhamodi, line 41: One thing we talked about yesterday was allowing this to be a toggleable field from the tutorial, and allowing it to persist throughout the rest of the escape room experience.

I do foresee issues around if a user refreshes on a scene room or something similar, that context is lost. Unless we store in cookies or something similar.

I am wondering what you think about if we should even consider this solution or just find a majority preference and go with that only.

jaydhulia: > I do foresee issues around if a user refreshes on a scene room or something similar, that context is lost. 

This shouldn't be an issue, we can store user preferences in the frontend using local storage and pure JS.

I think the user preferences way is definitely doable, the hardest part will just be building an Aframe button that is togglable. But even that shouldn't be too hard imo.

PR #147, opened 2021-07-17
Tutorial integration

jaydhulia:
Improve the tutorial flow.
Loom video: https://www.loom.com/share/ae5cee8690b944578b60fd5cbd0dc3c7


app/public/static/js/components/tutorial-instructions.js
```diff
@@ -0,0 +1,101 @@
...
+        const tutorial_keypress_transition_data = {
+          data: [
+            {
+              text:
+                "Great job! You can also pan around the room by clicking and dragging your mouse. During the escape room, you will have to interact with various objects to solve puzzles or get clues. To interact, click on the textbook (on teacher's desk) with your left mouse button. Give it a try!",
+            },
+          ],
+          currPosition: 0,
+        };
+        tutorial_keypress_transition_data["isTransition"] = false;
```
drbalar, line 59: can we define `isTransition` inside on line 58 similar to how we define `data` and `currPosition`? is there a reason to do it this way?

app/public/static/js/components/tutorial-instructions.js
```diff
@@ -0,0 +1,101 @@
...
+    this.textLabel.setAttribute("position", { x: 0, y: 0, z: -1 });
+    this.textLabel.setAttribute("width", 1);
+
+    el.appendChild(this.textLabel);
+    window.addEventListener("keydown", this.keyDownHandle);
+  },
+  keyDownHandle: function (e) {
+    const buttons = this.buttons;
+    const dir = this.dir;
+    if (e.key === buttons[this.cur_idx]) {
```
drbalar, line 37: can we also add a check for the CAPS WASD? If someone has caps lock on then they would not get past the first part i think

drbalar (changes requested):
looks good! I have just 2 things

app/public/static/js/components/tutorial-instructions.js
jaydhulia: Updated PR, no reason to do it that way, it was just a copy pasta

app/public/static/js/components/tutorial-instructions.js
jaydhulia: Updated, by forcing e.key to be lower case.

drbalar (approved):
Looks good! 🚀

drbalar (dismissed):
works nicely for me :D

ahmedhamodi (commented):
Overall looks 🔥 

Minor nit around the movement instructions to the popup -- can we add a slight delay from the moment the user presses "d" to move right so they can physically see themselves move right before the popup comes? Feels too abrupt at the moment.

aaronabraham311 (dismissed):
Approving for product and design

ahmedhamodi (approved):

PR #149, opened 2021-07-21
Add screenshot_url field to scenario schema to support frontend PR

ahmedhamodi:
*NEW* Loom demo: https://www.loom.com/share/7002cf944d8945b083a511569e9a1ffb

Associated frontend PR: https://github.com/uwblueprint/cdc-frontend/pull/105


app/library/postgres.py
```diff
@@ -224,6 +224,25 @@ async def get_scenario_by_friendly_name_from_postgres(
 
 
 async def delete_scenario_from_postgres(scenario_id: str, session):
+    # Get scene first
```
jaydhulia, line 227: This should be scenario, nice copy pasta 😛

app/library/postgres.py
```diff
@@ -224,6 +224,25 @@ async def get_scenario_by_friendly_name_from_postgres(
...
+    scenario_obj: Scenario = get_scenario(scenario_id, session)
+    if scenario_obj is None:
+        raise ValueError("Scenario ID not valid")
+
+    # Delete the screenshot of scenario, if it has one
+    if scenario_obj.screenshot_url:
```
jaydhulia, line 233: IMO `screenshot_url` is a weird name for this. It makes sense for the scene since we are taking a screenshot of the scene and storing it, but in this case it's not really a screenshot. `display_image_url` or something along those lines might be better.

app/library/postgres.py
ahmedhamodi: shoot LOL

app/library/postgres.py
```diff
@@ -224,6 +224,25 @@ async def get_scenario_by_friendly_name_from_postgres(
...
+        # Only delete from AWS if config is enabled for AWS support
+        if "aws" in config.get("app-env"):
+            s3_client = boto3.client(
+                "s3",
+                endpoint_url=f"https://s3.{config.get('s3.region')}.amazonaws.com",
+                region_name=config.get("s3.region"),
+            )
+            s3_client.delete_object(
+                Bucket=config.get("s3.bucket_name"),
+                Key=scenario_obj.screenshot_url,
```
jaydhulia, line 243: This won't work, the screenshot_url being saved is the full URL, and what you need to delete is only the key of the object. Try looking at `scene` table's screenshot_url vs the `scenario`:
`scene`: images/247bca4e5f3946619e2633243d9f90f2.jpeg
`scenario`: https://dcc-bp-public-dev.s3.us-east-2.amazonaws.com/assets/cd231a6d3603410a908633c01c7065a5

jaydhulia (changes requested):

app/library/postgres.py
ahmedhamodi: Yeah I noticed in the scene logic it was appending the asset prefix on the backend and not frontend. I guess this is why. Will fix that.

jaydhulia: Yeah, when I implemented that endpoint initially (the screenshot), FE didn't have that asset prefix variable. Now it can be prepended in either FE or BE.

ahmedhamodi: Properly set now :)
image

jaydhulia (approved):

ahmedhamodi (commented):
I'll merge this once frontend PR is approved:
https://github.com/uwblueprint/cdc-frontend/pull/105

@jaydhulia when you get a chance!

PR #150, opened 2021-07-22
Dhruvin room 3

drbalar:
Loom: https://www.loom.com/share/b4fd8e7ce491473fa727e3d134e812c9

Note: The images and object files are local. Not on S3. If everything from design side is good then I'll upload to S3 and update this PR.


aaronabraham311 (commented):
Nice work Dhruvin! Looks really good. 

Couple of comments:
- I think you forgot the body text for the jigsaw puzzle. "A piece of paper fell out of my guitar. Interesting"
- We also need some context text for the plant pane. Can you add another pane to the front saying: "What's this piece of paper doing underneath the pot? I'm always misplacing my things when I get anxious"
- Can we put other messages in the phone screenshot that are not related to the solution? For example, we could have another message from a person saying "Don't forget, card game at 9 tonight"

Overall, do you feel that this is challenging and interesting? If not, we can revisit the choices we made for the room

ahmedhamodi:
"smell the plants, sniff sniff" why are you so funny Dhruvin 😂

ahmedhamodi (commented):
Are there some unnecessary files in this PR?

ie. f5.jpg and f5.png? f3.jpg and f3.png? f9.jpg and f9.png?

drbalar:
> Are there some unnecessary files in this PR?
> 
> ie. f5.jpg and f5.png? f3.jpg and f3.png? f9.jpg and f9.png?

Yeah those will be removed once everything uploaded to S3. I just pushed those for now if anyone wants to run it locally and try it out

drbalar:
> * I think you forgot the body text for the jigsaw puzzle. "A piece of paper fell out of my guitar. Interesting"
image



> * We also need some context text for the plant pane. Can you add another pane to the front saying: "What's this piece of paper doing underneath the pot? I'm always misplacing my things when I get anxious"
image



> * Can we put other messages in the phone screenshot that are not related to the solution? For example, we could have another message from a person saying "Don't forget, card game at 9 tonight"
image


> * Is it possible to add some padding on the top for the title text on modals? It's too close to the top
image


> Overall, do you feel that this is challenging and interesting? If not, we can revisit the choices we made for the room

It is challenging in my opinion.. maybe too challenging? Like having extra image frames that are not relevant, that makes it a little harder but that might be good in a way.

It's hard to say without bias if it was interesting or boring because I knew the steps and answers already once it was built. Pretty neutral on that.

aaronabraham311 (dismissed):
Approving for product and design.

We can change the difficulty after our user testing session if people are having a really hard time.

jaydhulia (approved):
Tested it using local config and AWS config, works great on both. In terms of difficulty, I was able to solve it without too much effort, so I think it's solvable. I really like it!!

PR #151, opened 2021-07-23
reverse ratio for jigsaw puzzle

drbalar:


app/public/static/js/components/ordered-puzzle.js
```diff
@@ -84,9 +84,16 @@ AFRAME.registerComponent("ordered-puzzle", {
         const puzzlePiece = puzzlePieceCache[index];
 
         // Calculate the puzzle piece width and height based on image ratio
-        const ratio = rawImageEl.width / rawImageEl.height;
+        let ratio = rawImageEl.height / rawImageEl.width;
```
jaydhulia, line 87: This should stay as `rawImageEl.width / rawImageEl.height;` right? Since for non-jigsaw, we want the current behaviour.

drbalar: I think currently, even for non-jigsaw the issue of squishing seems to exist. When i changed the ordered puzzle to use a narrow picture, you can see the effect is still there
image

jaydhulia: Gotcha, okay, code perspective, it looks good to me. @amoliksingh can you test out with the images you were trying with the ordered puzzle and make sure it still spawns the visual targets okay?

ahmedhamodi (commented):
Hm I do find it very strange conceptually that for jigsaw we want ratio to be `width / height` but for ordered puzzle we want `height / width`. Seems like we may be misusing the fields width and height or are doing our scaling wrong. But if this works, I'm okay with it!

jaydhulia (approved):

drbalar:
> Hm I do find it very strange conceptually that for jigsaw we want ratio to be `width / height` but for ordered puzzle we want `height / width`. Seems like we may be misusing the fields width and height or are doing our scaling wrong. But if this works, I'm okay with it!

It is mainly for what we want to limit our restrict our image to. For ordered puzzles, its a long puzzle and so need to limit the widths of the images. For the jigsaw, the width is not a big issue, but the vertical height needs to stay at a specific size so we have to reverse it

PR #153, opened 2021-07-24
Aframe Button integration

jaydhulia:
https://www.loom.com/share/77a2947d003143cf8656a032f542a4a3


ahmedhamodi (commented):
I think this was mentioned before by Aaron, but the placements of the buttons appear off? The X is too close to the top and the next/prev buttons should be closer to the bottom of the pane. It feels like it is all shifted up by 10 or so pixels. Is that something we can fix in this PR?

jaydhulia:
@ahmedhamodi 
That is related to the font change that we did and I would say that’s not in the scope of this ticket. This one of just for button integration <-> hover state. When we fix the text positioning, we need to fix up all the texts in all the popups, which includes stuff like blackboard title etc., so that should be a diff PR.

ahmedhamodi (approved):
Code approval. Pending design approval.

drbalar (approved):
Looks good, and works good for me locally too

aaronabraham311 (commented):
In the ticket, we wanted button hover to have a border. Is that not possible?

jaydhulia:
@aaronabraham311 yeah, it seems that adding a border in Aframe isn’t possible without a massive dev effort, and after talking with Jack, we concluded that a change of colour should be sufficient.

aaronabraham311 (approved):

PR #159, opened 2021-08-04
Fix X Button Position on Modal

vivianliu0:
Image of text pane:
image


aaronabraham311 (approved):

ahmedhamodi (commented):
I might be being too picky but it seems too low? The distance from top > distance from left side 😅 I feel like 6.3 or 6.35 might be better.

vivianliu0:
> I might be being too picky but it seems too low? The distance from top > distance from left side 😅 I feel like 6.3 or 6.35 might be better.

Sure I can update this, I noticed that the spacing of the Prev and Next button was greater for the bottom margin so I kept it consistent. I'll change them to all have the same distance from the edges

vivianliu0:
> > I might be being too picky but it seems too low? The distance from top > distance from left side 😅 I feel like 6.3 or 6.35 might be better.
> 
> Sure I can update this, I noticed that the spacing of the Prev and Next button was greater for the bottom margin so I kept it consistent. I'll change them to all have the same distance from the edges

Updated image in description which reflects new changes

jaydhulia (approved):
LGTM!

ahmedhamodi (approved):

PR #162, opened 2021-08-11
Update contents for room 2 

jaydhulia:
- Update frame + asset
- Update hints
- Update jigsaw puzzle
- Make password case insensitive (on all rooms)

https://www.loom.com/share/e57dcbb0cffd43b0a6ea85fd333ffc3c


samples/jigsaw_puzzle/jigsaw-prototype.py
```diff
@@ -46,8 +47,14 @@
...
         s3_key_counter += 1
+        filenames.append(
+            "https://dcc-bp-public-dev.s3.us-east-2.amazonaws.com/"
+            + s3_key_prefix
+            + str(s3_key_counter)
+            + ".png"
+        )
 
-
+print(filenames)
```
aaronabraham311, line 57: Remove?

aaronabraham311 (approved):
Product approval

samples/jigsaw_puzzle/jigsaw-prototype.py
jaydhulia: It is needed for jigsaw puzzles, the script isn't part of the main program, but a script we (devs) can use to quickly get a jigsaw onto s3 and get the filenames :)

app/public/static/js/components/keypad.js
```diff
@@ -38,7 +38,7 @@ AFRAME.registerComponent("keypad", {
     }
 
     el.addEventListener("superkeyboardinput", function (event) {
-      if (event.detail.value === password) {
+      if (event.detail.value.toLowerCase() === password.toLowerCase()) {
```
ahmedhamodi, line 41: Todo for later: allow admin to specific whether they want case sensitive or not -- case sensitivity might be important for some custom admin created puzzles!

samples/jigsaw_puzzle/jigsaw-prototype.py
```diff
@@ -7,7 +7,7 @@
 from botocore.config import Config
 from numpy import frombuffer, uint8
 
-f = open("b64/lightOn", "r")
+f = open("b64/viv", "r")
```
ahmedhamodi, line 10: Is this intentional?

jaydhulia: Yeah, it doesn't matter. It is a script, the filename changes based on what we are doing. It was updated to viv last since she requested a file to be uploaded to s3 through jigsaw and I used her name as filename

aaronabraham311 (approved):

PR #165, opened 2021-08-11
Dhruvin room 4

drbalar:
Loom: https://www.loom.com/share/1d6b7942860a430686b82a68d335c275

I did not push the assets to this PR since they will be renamed/removed eventually.
Like last time, once approved in design terms, i will add assets and images to s3 and update PR


aaronabraham311 (dismissed):
Product approval

ahmedhamodi:
The confetti 😂 

No comments from me, will leave code review for Jay when he gets a chance. I really like the room!

jaydhulia:
Not much to review for the code here tbh, let's just get all the assets into S3, merge this in, and test in production 😄

jaydhulia (changes requested):
@drbalar the insert script doesn't work after the latest merge from the master, because there were some assets moving around / deleted. So doing `make install` now throws an error because assets aren't in created with the expected ID. Do you mind fixing that? 
And then I can test it out locally. Thanks

drbalar:
@jaydhulia Fixed the IDs to match the new data. Should work now, let me know if it does not still for you

jaydhulia (approved):
LGTM!

PR #167, opened 2021-09-25
Delete puzzle assets from S3

jaydhulia:
- When an object changes from interactable -> non-interactable, OR it's puzzle type changes, we should clean up the old assets from S3 related to the old object


app/library/postgres.py
```diff
@@ -441,3 +441,54 @@ async def delete_object_in_postgres(scene_id: str, object_id: str, session):
...
+    del old["asset_details"]
+    del new["asset_details"]
+    old_obj = Object(**old)
+    new_obj = Object(**new)
+    if not old_obj.is_interactable:
+        # old object wasn't intractable, nothing to clean up, return
+        return
+
+    old_puzzle_type = old_obj.animations_json["blackboardData"]["componentType"]
+    new_puzzle_type = "NA"
```
drbalar, line 458: Is there a reason to have it `NA` specifically?

drbalar (approved):
looks good codewise. just one comment

PR #171, opened 2021-10-13
Dhruvin 404 page

drbalar:


app/routes/base.py
```diff
@@ -106,14 +106,10 @@ class NotFoundHandler(tornado.web.RequestHandler):
...
 
-    async def prepare(self):
-        self.set_header("Content-Type", "application/problem+json")
-        title = "Not Found"
-        message = "Invalid Path"
-        status_code = 404
-        self.set_status(status_code)
-        response_error = {"status": status_code, "title": title, "message": message}
-        await self.finish(response_error)
+    def write_error(self, status_code: int, **kwargs: Any) -> None:
```
jaydhulia, line 109: Ah, so the `NotFoundHandler` is the default handler for all non-existent routes (https://github.com/uwblueprint/cdc-backend/blob/master/app/__main__.py#L119). The reason why I used `prepare` is because `prepare` is run by Tornado for ALL requests. Therefore, it would run for all requests that don't match any of the other handlers. This is the behaviour we want, so we should leave prepare as it is.
Instead, where you want to override this is here: https://github.com/uwblueprint/cdc-backend/blob/master/app/routes/ui/landing_page.py#L37 (the UI handler 404 page for when scenarios aren't found)

drbalar: image
I did this but nothing happens now. I just run into this simple json error page:
image

Is something wrong in my code?

drbalar: YEP!

jaydhulia (approved):

PR #172, opened 2021-10-16
Update instructions for docker

jaydhulia:


drbalar (approved):

docs/dev-deployment-README.md
```diff
@@ -45,3 +45,30 @@ Make sure you have sufficient AWS permissions for the actions required. If you h
...
    - Tail error logs - `tail -f /var/log/nginx/error.log`
+
+## AWS EC2 - Docker
+
+1. Create an AWS EC2 instance (ask Jay for more details). Make sure the public IP is selected, or else it won't be accessible from outside. When choosing the instance type, choose "Ubuntu 20.04".
+1. Change the security group rules to allow Inbound traffic TCP to Port 443 from `Anywhere`
+1. Download the latest required dependencies:
+   - `sudo apt-get update`
+   - `sudo apt-get install docker.io unzip`
+   - `curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"`
```
ahmedhamodi, line 56: To be honest a lot of this is over my head as I have very little docker experience, but as a sanity check is this good for both Windows and Mac? Reason I ask is because "linux" is in the string so whenever I see that I think non-Windows support 😆

ahmedhamodi: Ah thanks for explanation!

ahmedhamodi (approved):

PR #173, opened 2021-10-20
Landing page implementation

jaydhulia:


ahmedhamodi (approved):
Works great. Back to home button on 404 works too, so that is great!

PR #181, opened 2021-10-30
Complete scene 5

ahmedhamodi:


scripts/mockdata/scenes_e1s5.json
```diff
@@ -0,0 +1,28 @@
+[
+  {
+    "name": "Bedroom with Cat",
```
ahmedhamodi, line 3:
```suggestion
    "name": "Bedroom with Cat - 3",
```

scripts/mockdata/scenarios_e1.json
```diff
@@ -97,6 +97,28 @@
           }
```
drbalar, line 106: I think we're missing a few slides for the scene intro here 
image

drbalar (approved):
Looks good 🚀

scripts/mockdata/scenarios_e1.json
ahmedhamodi: Added 👌

PR #185, opened 2021-11-17
Dhruvin small things

drbalar:
- new girl asset
- disable computer puzzle in scenes 4 and 5
- replace scene 4 final puzzle to be jigsaw
- disable keyboard close in tutorial

image

image


jaydhulia (approved):

ahmedhamodi (commented):
I like the new jigsaw puzzle image! The only issue is the order of the positions, since each chair is by itself in a column, it can be in multiple different combinations. perhaps adding a number to the positions (so like #1 below chair 1, #2 below char 2, etc) would help?

drbalar:
I didnt add anything thinking there is a slight distinction since the rightmost chair's back is cutoff and the leftmost person's knees are cutoff from the picture. But I guess that can still lead to some confusion since the jigsaw pieces are slightly spaced out.

I'll try adding the numbers

ahmedhamodi:
Yeah, maybe beside the X and checkmark? A little number to indicate order? Not sure. Good point on the knee + chair being cut out, my brain didn't even process that. But given that, I would assume others would also not process that haha

PR #187, opened 2021-11-27
Add a solution panel

jaydhulia:
- [x] Add a spoiler panel before the last hint in every scene by default
- [x] Add a solution as the last hint for every scene

image

image


drbalar (approved):
Looks good

PR #188, opened 2021-12-02
Change landing page verbiage

ahmedhamodi:
Request from client:
- remove plural since only one room
- change verbiage of room to be just the content in the brackets

Before:
image

After:
image

### DIFF
--- a/app/public/templates/home.html
+++ b/app/public/templates/home.html
@@ -50,7 +50,7 @@
       </nav>
       {% if len(res_scenarios) == 0 %}
       <h2>No escape rooms found</h2>
-      {% else %} {% if len(res_scenarios) > 1 %}
+      {% elif len(res_scenarios) > 1 %}
       <h2>Try out one of the escape rooms below!</h2>
       {% else %}
       <h2>Try out the escape room below!</h2>
@@ -64,7 +64,7 @@
           >
         </div>
       </div>
-      {% end %} {% end %}
+      {% end %}
     </div>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


app/public/templates/home.html
```diff
@@ -60,9 +64,7 @@ <h5 class="card-title">{{ scenario['title'] }}</h5>
           >
         </div>
       </div>
-      {% end %} {% else %}
-      <h1>No escape rooms found</h1>
-      {% end %}
+      {% end %} {% end %}
```
jaydhulia, line 67: There's an extra dangling `end` here.

app/public/templates/home.html
```diff
@@ -48,9 +48,13 @@
           alt=""
         />
       </nav>
-      {% if len(res_scenarios) > 0 %}
-      <h1>Try out one of the escape rooms below!</h1>
-      {% for scenario in res_scenarios %}
+      {% if len(res_scenarios) == 0 %}
+      <h2>No escape rooms found</h2>
+      {% else %} {% if len(res_scenarios) > 1 %}
```
jaydhulia, line 53: This doesn't quite have the effect you want. In tornado templates, the else if statements are `elif`. So you want something along the lines of:
```
{% if *condition* %}...{% elif *condition* %}...{% else %}...{% end %}
```

jaydhulia (changes requested):

app/public/templates/home.html
ahmedhamodi: Did not know that! Thank you. I tried google for the else if condition but people were just doing nested if's 😅

jaydhulia (approved):

jaydhulia (approved):

PR #193, opened 2021-12-21
Hash scene numbers

jaydhulia:
- Hashes the scene number (NOT for security purposes, but rather just to avoid people randomly skipping to the next scene... if they can figure out the hash then they can also read source code 😄  so they aren't the primary audience for this change). 
- First 3 and last 3 characters are always randomly generated for the hashes
- Scene "0" is a special case


ahmedhamodi (commented):
A nit, but is it possible to have the room lengths be the same?

For example room 0 is: http://localhost:8888/youth-in-crisis/3500d0b
But room 1 is: http://localhost:8888/youth-in-crisis/f7a6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b531

If we can make them all the same, that would be ideal imo

Otherwise this works well.

ahmedhamodi:
Thinking about this more, we honestly don't even need the hash. Just random 3 chars before, the actual room number, then random 3 after. Strip the 6 chars to get the actual room number.

ie. 3500d0b for room 0, xxx1xxx for room 1, xxx2xxx for room 2, etc. Makes this logic much more simple right?

jaydhulia:
@ahmedhamodi a couple of things there:
1. For the inconsistency, we will always have inconsistent length, even if we do it that way, as room number 10 will always be 1 length bigger than 0-9. 
2. Yes that might be simpler logic, but I have already written this PR up and tested it 😄 I don't want to test it again for all the edge cases for a nit.
I think mainly what I am getting at is, no one should be looking at the URL other than the friendly URL they start with. So as long as the functionality works, let's not waste too much time on the details of the implementation itself.

jaydhulia:
@ahmedhamodi I can make room 0 the same length if that helps though, let me know and I can fix that part.

ahmedhamodi:
Yeah that is fair, let's leave it for now. A todo for later perhaps :)

ahmedhamodi (approved):
Works locally!

drbalar (approved):
Looks good, my only question is do we plan on making the `tutorial` and `completed` pages also "hashed" or leave those as is?

ahmedhamodi:
No need to hash tutorial imo. Completed has some merit, but I don't think it's needed.

PR #197, opened 2022-01-20
Small changes to readme and home.html template

ahmedhamodi:


app/public/templates/home.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>ConnecTeen Interactive</title>
+    <title>Houdini - ConnecTeen Interactive</title>
```
ahmedhamodi, line 4: I made this same change on the cdc-frontend repo: https://github.com/uwblueprint/cdc-frontend/pull/138/files#diff-c2cc24bc9001b11b6add48a4cd8f893d5d6c6e4d1bd254158bd14ab997f552cdR27

Let me know if I should revert in cdc-frontend or if the change makes sense to do here too.

jaydhulia: Let's keep it the other way around for here. Since it is the student env, we want ConnecTeen to be highlighted more :)

ahmedhamodi: Reverted!

ahmedhamodi:
Merging because only a readme change now lol

PR #199, opened 2022-01-27
WIP: Escape room 2 - Scene 1

drbalar:


ahmedhamodi (commented):
There's an `untitled.glb` file -- is that intentional?

image

PR #202, opened 2022-02-10
Add default properties for each background

jaydhulia:
Add default camera properties, position, rotation, scale and hint for each of our background assets

### DIFF
--- a/configs/dev-aws-config.yaml
+++ b/configs/dev-aws-config.yaml
@@ -279,4 +279,4 @@ scene_data:
       - 0.425
   hints:
     - "You can use your mouse to pan around, and use the arrow keys / WASD to move around"
-    - "If you are seeing this page and aat aren't the admin, please contact Calgary ConnecTeen to let them know that a solution is missing :)"
+    - "If you are seeing this page and aren't the admin, please contact Calgary ConnecTeen to let them know that a solution is missing :)"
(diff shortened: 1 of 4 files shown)


configs/dev-aws-config.yaml
```diff
@@ -131,3 +131,152 @@ scene_data:
...
+      - 0.425
+      - 0.425
+      - 0.425
+    43:
+      - 0.425
+      - 0.425
+      - 0.425
+  hints:
+    - "You can use your mouse to pan around, and use the arrow keys / WASD to move around"
+    - "If you are seeing this page and aat a ren't the admin, please contact Calgary ConnecTeen to let them know that a solution is missing :)"
```
ahmedhamodi, line 282: Typo :)

configs/dev-config.yaml
```diff
@@ -130,3 +130,152 @@ scene_data:
...
+      - 0.425
+      - 0.425
+      - 0.425
+    43:
+      - 0.425
+      - 0.425
+      - 0.425
+  hints:
+    - "You can use your mouse to pan around, and use the arrow keys / WASD to move around"
+    - "If you are seeing this page and aat a ren't the admin, please contact Calgary ConnecTeen to let them know that a solution is missing :)"
```
ahmedhamodi, line 281: Same here

ahmedhamodi (commented):
Small typo. Otherwise code looks good to me. I'll test it later today if Dhruvin doesn't beat me to it.

configs/dev-aws-config.yaml
jaydhulia: Fixed!

configs/dev-config.yaml
jaydhulia: Fixed!

ahmedhamodi: I think there is still a typo. `...page and aat aren't the...`. The "aat" shouldn't be there.

ahmedhamodi (dismissed):
One you remove the `aat` text from both the configs, this looks good to me.

configs/dev-config.yaml
jaydhulia: Updated

ahmedhamodi (approved):

PR #204, opened 2022-02-15
Small typo fixes

ahmedhamodi:


ahmedhamodi (commented):
As per Jay's suggestion, all changes here are made manually on the deployed version of the backend. However, we should still merge this eventually but no rush.

cc: @drbalar @jaydhulia

PR #208, opened 2022-03-02
March 2 Bug fixes from bug bash

jaydhulia:


app/library/api_schemas.py
```diff
@@ -19,11 +22,14 @@
 admin_scenario_post_handler_schema = {
     "type": "object",
     "properties": {
-        "name": {"type": "string", "pattern": r"^[\'a-zA-Z0-9 _-]{1,50}$"},
+        "name": {
+            "type": "string",
+            "pattern": r"^[\:\(\)\'\?\!\.\",a-zA-Z0-9 _-]{1,50}$",
```
drbalar, line 27: Should this be 2000 instead of 50 to match room description size?

image

drbalar (approved):

